PR #205, opened 2014-08-11
CA-141409 strict nfs version

matelakat:
Amended the original patches for xs64bit


robertbreker:
This pull request can be closed.
While the patch would not hurt, it is not essential for xs64bit. We can avoid the churn at this stage.

PR #206, opened 2014-08-11
vhd-meta.py

germanop:
When we moved to github we dropped many utilities just because they were
not packaged in the RPM.

vhd-meta.py is one of those utilities.

The commits should be kept (IMHO) separated for historical reasons.
The first one is the original one from Andrei they we dropped.

The second one is my improved version to work with LVHD SRs (the previous one
was meant only for file based SRs). It was already reviewed and ready to be committed but
after we moved, because the original was not there the patch was dropped.

The third one is an improved version from Keith to cope with nasty situations were the footers
were corrupted. I did not spend time reviewind the patch in full so there is room for improving
the commit message.

These 3 patches are the state of the art of this script. There many versions scattered throughout
tickets from customers. An old version has been discarded because it was supersed by the first
commit from Andrei (it was a first attempt from Andrei to support a customer but then he refined
the patch).
Another version is not included because it was a hybrid: it had the second commit on top of a very
old version (see above).

So, these 3 patches together are exhaustive.

### DIFF
--- a/tools/vhd-meta.py
+++ b/tools/vhd-meta.py
@@ -1,7 +1,22 @@
 #!/usr/bin/python
-
-# Script for collecting VHD metadata for a given VHD chain (e.g. from a remote 
-# site), and for recreating the full VHD files (with zeros written to data 
+#
+# Copyright (C) Citrix Systems Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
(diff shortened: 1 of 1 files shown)


BobBall:
Just a note for improvement... The script appears to assume that each VDI has the same BAT size.  Rackspace had to modify it to extract the metadata.

germanop:
In out case the BAT is always the same.

If they want to send the patch over, we can include it here

PR #220, opened 2014-09-09
CA-145177: Cleanup flock.py module

matelakat:


germanop:
@matelakat any reason why you closed it?

PR #221, opened 2014-09-17
CA-146138: don't refresh slaves for local SRs

tmakatos:
Signed-off-by: Thanos Makatos <thanos.makatos@citrix.com>


germanop:
Code merged in xs64bit and then in master with merge 7b9e693d

PR #227, opened 2014-10-02
CA-129230: Output of SM is truncated in logs (by syslog)

kostaslamda:


germanop:
Code in master via 56e6aca1 and 2ee924ac

PR #237, opened 2014-12-17
Improving readcache check for file-based SRs

germanop:
The check introduced with commit 4be47e0 happens in the
load function.
This means that it has to be done for the load function
of every file based SR.

Putting the check in the FileSR constructor will make
it available for every derived class, without the need
to add the check for every new SR (as it happened to
me while playing with OCFSSR)

Signed-off-by: Germano Percossi <germano.percossi@citrix.com>


germanop:
This PR is under testing in our internal patch queue

PR #238, opened 2014-12-19
CA-128095

kostaslamda:


germanop:

germanop:
OK, back on this after quite some time :-) 

The raise in there is too drastic.
We need just to fix the case where we leave the (huge and horrible) try block.
There are too many things that can fail in there, for different reasons and not all of them should
make the entire load fail.

Just check that at the end of the block we have at least one session and if not, raise a generic iscsi failure.

I think this is the simplest and lest invasive thing we can do

drivers/LVHDoISCSISR.py
```diff
@@ -153,6 +153,7 @@ def load(self, sr_uuid):
                     self.session.xenapi.PBD.set_device_config(pbd, dconf)
             except:
                 util.logException("LVHDoISCSISR.load")
+                raise
```
germanop: raising here has many unwanted effects.

Let's just raise if we did not manage to have any valid session (the reason why the next indexing fails).

MarkSymsCtx:
Superseded by PR-358

PR #241, opened 2015-02-01
CA-159457: Fix parsing in sg_readcap to handle READ CAPACITY (16) output

robertbreker:
Please review, comment and merge if OK.

### DIFF
--- a/drivers/scsiutil.py
+++ b/drivers/scsiutil.py
@@ -644,6 +644,10 @@ def sg_readcap(device):
         # retry one time for "Capacity data has changed"
         (rc,stdout,stderr) = util.doexec(readcapcommand)
     if rc != 0:
-        raise util.SMException("util.sg_readcap(%s) failed" % (device))
-    (blockcount,blocksize) = stdout.split()
+        raise util.SMException("scsiutil.sg_readcap(%s) failed" % (device))
+    match = re.search('(^|.*\n)(0x[0-9a-fA-F]+) (0x[0-9a-fA-F]+)\n$', stdout)
+    if not match:
+        raise util.SMException("scsiutil.sg_readcap(%s) failed to parse: %s"
+                               % (device, stdout))
+    (blockcount, blocksize) = match.group(2, 3)
     return (int(blockcount, 0) * int(blocksize, 0))
--- a/tests/test_scsiutil.py
+++ b/tests/test_scsiutil.py
@@ -1,0 +1,32 @@
+import unittest
+import mock
+
+import scsiutil
+
+
+class Test_sg_readcap(unittest.TestCase):
+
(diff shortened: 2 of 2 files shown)


robertbreker:
(The Travis CI build failed due to an upstream compatibility issue with python 2.6 that should be fixed in Astroid 1.3.5 - https://bitbucket.org/logilab/astroid/commits/947faa7b39ed)

drivers/scsiutil.py
```diff
@@ -644,6 +644,11 @@ def sg_readcap(device):
...
     if rc != 0:
-        raise util.SMException("util.sg_readcap(%s) failed" % (device))
-    (blockcount,blocksize) = stdout.split()
+        raise util.SMException("scsiutil.sg_readcap(%s) failed" % (device))
+    match = re.search('(^|.*\n)(0x[0-9a-fA-F]+) (0x[0-9a-fA-F]+)\n$', stdout)
+    if not match:
+        raise util.SMException("scsiutil.sg_readcap(%s) failed to parse: %s"
+                               % (device, stdout))
+    blockcount = match.group(2)
+    blocksize = match.group(3)
```
siddharthv: @robertbreker Could we shorten this to something like below ?
`(blockcount, blocksize) = match.group(2, 3)`

tests/test_scsiutil.py
```diff
@@ -0,0 +1,31 @@
...
+        doexec.return_value = (0, fake_out, '')
+        self.verify_sg_readcap(doexec, 500074307584)
+
+    @mock.patch('util.doexec')
+    def test_capacity_data_changed_rc6(self, doexec):
+        fake_out = "0x3a376030 0x200\n"
+        doexec.side_effect = [(6, 'something else', ''), (0, fake_out, '')]
+        self.verify_sg_readcap(doexec, 500074307584)
+
+
```
siddharthv: Can we get rid of this extra line ?

robertbreker: Certainly (pep8)

drivers/scsiutil.py
robertbreker: Certainly; shouldn't make a difference.

tests/test_scsiutil.py
siddharthv: Yep!

siddharthv:
@robertbreker Looks good now. Thanks!!

germanop:
Sid, 2 things here:
* Travis is failing but if you know the reason and it is ok for you I have no objections
* The second commit message starts with [ ]: for "git am" everything starting this way is treated as not part of the commit. Consider while including this patch to either add the CA ticket first (then you
can keep the brackets) or just remove the brackets. So far we have always been using the former.

Once this 2 points are sorted, you are good to go.

PR #242, opened 2015-02-02
CP-10986: Return read-caching status on VDI.attach.

siddharthv:

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -977,7 +977,28 @@ class VDI(object):
         self._vdi_uuid   = uuid
         self._session    = target.session
         self.xenstore_data = scsiutil.update_XS_SCSIdata(uuid,scsiutil.gen_synthetic_page_data(uuid))
+        self.__o_direct  = None
         self.lock        = Lock("vdi", uuid)
+
+    def get_o_direct_capability(self, options = {}):
+        """Returns True/False based on licensing and caching_params"""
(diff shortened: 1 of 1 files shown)


drivers/blktap2.py
```diff
@@ -1467,6 +1467,7 @@ def attach(self, sr_uuid, vdi_uuid, writable, activate = False):
         # Return backend/ link
         back_path = self.BackendLink.from_uuid(sr_uuid, vdi_uuid).path()
         struct = { 'params': back_path,
+                   'o_direct': util.read_caching_is_restricted(self._session),
```
simonjbeaumont: I think this is a race condition. The blktap driver gets hold of this value here and it's this that's passed to the Tapdisk class: https://github.com/siddharthv/sm/blob/CP-10986/drivers/blktap2.py#L1564-L1565.

In between these two places the licensing could have changed meaning the tapdisk is forked with one option but the return data says the opposite.

simonjbeaumont:
Also, we need to ensure that all implementations of the SMAPI returns a sensible default for all SRs. I've just grep'd around and see a few other places where this kind of dictionary is created (obviously without `o_direct`) and returned.

siddharthv:
@simonjbeaumont Ping!

franciozzy:
The patch looks good to me. Ship it!

Reviewed-by: Felipe Franciosi <felipe.franciosi@citrix.com>

PR #243, opened 2015-02-04
CAR-1442: Enable no-O_DIRECT read caching for VDIs on EXT and NFS SRs

siddharthv:
Pass the "-D" argument to tap-ctl when opening an EXTFileVDI or an NFSFileVDI
to allow read caching by removing O_DIRECT.

Allow disabling read caching per-SR by setting other-config:o_direct to true.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>
Reviewed-by: Thanos Makatos <thanos.makatos@citrix.com>
Imported-by: Siddharth Vinothkumar <siddharth.vinothkumar@citrix.com>

(cherry picked from commit 4be47e0b137652ef4799931a87bca6ea4ef0dd67)

Fixed missing code chunks on master.

Conflicts:
	drivers/blktap2.py

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -395,6 +395,8 @@ class TapCtl(object):
         if options.get("timeout"):
             args.append("-t")
             args.append(str(options["timeout"]))
+        if not options.get("o_direct", True):
+            args.append("-D")
         cls._pread(args)
 
     @classmethod


germanop:
I am not commenting on the code, I will let Thanos check if this is the missing hunk.
On the commit message side, though, I have few comments.
We need the commit message to understand what is going on, so:

* writing "cherry picked" is not correct because this is one hunk from a very long patch.
* What's the point of keeping the "Conflict" part
* Given this is not a cherry pick, the Imported-by does not apply
* In general, if there are minor changes with the import (or cherry-pick) it is customary
  (see linux kernel doc) for the importer of the patch to add themselves to the list
  (imported-by or signed-off-by, I do not care) but with a line before in square brackets
  explaining the changes), something like:

```
Signed-off-by: <original author>
[JohnDoe modified this and this]
Signed-off-by: John Doe <john@doe.org>
```

In this case, given it is a hunk forgotten from a previous commit (grrrrrr), I would suggest
to keep Ross as the author and his Signed-off-by tag, change the commit message (it does not make
sense any more if you compare what is written with the code) with the ticket number of
the bug due to this problem and write a commit message explaining what this code is coming
from.

siddharthv:
@tmakatos thanks :)
@germanop I've amended the commit message, could you please have another look ?

germanop:
The format of the commit is fine.
For the exact wording and who should be the author (you or Ross)
I'll leave it to @rosslagerwall

rosslagerwall:
This is OK with me. I don't mind who the author is, just do whatever is easier.

siddharthv:
@rosslagerwall thanks :)

germanop:
OK for me

PR #245, opened 2015-02-20
CA-159585: Fix broken travis-CI for SM

siddharthv:
As part of out travis use, we do a pylint check on new pull requests.
Recently, as of pylint(v1.4) support for python 2.6 has been dropped.
Also, pylint uses `astroid` as an external dependency who's changes have
also broken support for python2.6.

Since we use python 2.7 on trunk, we should drop testing our patches on
python 2.6 and switch to 2.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: python
 python:
-    - "2.6"
+    - "2.7"
 # command to install dependencies
 install:
-      - "pip install -r requirements.txt --use-mirrors"
+      - "pip install -r requirements.txt"
 # command to run tests
 script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: python
 python:
-    - "2.6"
+    - "2.7"
 # command to install dependencies
 install:
-      - "pip install -r requirements.txt --use-mirrors"
```
chandrikas: What does --user-mirrors do? Why don't we need it anymore?

chandrikas:
Looks good. Please merge to staging branch.

PR #247, opened 2015-02-27
CP-11278: Add supported nfs versions to sr-probe

siddharthv:
This patch adds the list of supported nfs versions for a given server
based on the 'nfsversion' key in device_config.

### DIFF
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -59,7 +59,7 @@ DRIVER_CONFIG = {"ATTACH_FROM_CONFIG_WITH_TAPDISK": True}
 PROBE_MOUNTPOINT = "probe"
 NFSPORT = 2049
 DEFAULT_TRANSPORT = "tcp"
-
+PROBEVERSION = 'probeversion'
 
 class NFSSR(FileSR.FileSR):
     """NFS file-based storage repository"""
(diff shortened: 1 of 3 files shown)


siddharthv:
@chandrikas I think we should use a new key in device_config, perhaps `nfsprobe` rather than relying on `nfsversion`.

chandrikas:
Yes, just a flag to indicate we want to probe for supported version. Why can't we return supported version by default BTW? What is the use case for that?

drivers/NFSSR.py
```diff
@@ -110,7 +110,10 @@ def validate_remotepath(self, scan):
 
     def check_server(self):
         try:
-            nfs.check_server_tcp(self.remoteserver, self.nfsversion)
+            if self.srcmd.params.get('device_config').has_key(PROBEVERSION):
```
chandrikas: Can we check self.dconf? That should already be populated with all device_config parameters.

drivers/NFSSR.py
```diff
@@ -118,9 +121,14 @@ def check_server(self):
 
     def mount(self, mountpoint, remotepath, timeout = 0):
         try:
-            nfs.soft_mount(
-                mountpoint, self.remoteserver, remotepath, self.transport,
-                timeout=timeout, nfsversion=self.nfsversion)
+            if self.srcmd.params.get('device_config').has_key(PROBEVERSION):
```
chandrikas: Can we check self.dconf? That should already be populated with all device_config parameters.

drivers/NFSSR.py
```diff
@@ -151,7 +159,7 @@ def probe(self):
 
         self.mount(temppath, self.remotepath)
         try:
-            return nfs.scan_srlist(temppath)
+            return nfs.scan_srlist(temppath, self.dconf.get('server'), self.srcmd.params.get('device_config'))
```
chandrikas: We could just pass self.dconf to nfs.scan_srlist here. That has all the parameters we need.

chandrikas:
@siddharthv , the patch looks good. Please change the commit message to reflect the device_config key we use to probe version (commit message still says nfsversion). Changes can be pushed once commit message is changed.

PR #249, opened 2015-03-27
Updates for CentOS 7

rosslagerwall:


germanop:
Hi Ross,

Are these patches *exactly* the same we have been testing internally?

rosslagerwall:
Yes.

germanop:
Ross, everything is ok but I have only one concern about
"Disable the default multipath udev rules" patch.

The rules names 40-multipath was named this way to override the rule with the same name
in CentOS,
I guess this rules has been numbered differently now, so the original sm one should have been
numbered accordinglym withouth any need to override it with a fake one.

germanop:
Acked-by: Germano Percossi <germano.percossi@citrix.com>

PR #251, opened 2015-04-09
CA-11936: Automatically plug block-based RAW VDIs via blkback

franciozzy:
This patch causes block-based (LV) RAW VDIs to be automatically plugged by
blkback. This is achieved as follows:

During VDI.attach, we detect that the SR handles LVM and that the VDI driver
is 'aio'. In that case, we write 'backend-kind: vbd' to xenstore_data, which
causes the toolstack to use the /local/domain/0/backend/vbd/ path instead of
.../vbd3/. This is the preferred method following CP-11511. By writing to
.../vbd/, blkback wakes up instead of tapback.

During VDI.activate, we create the /dev/sm/backend/ node using the maj/min
numbers of the actual VDI (LV) instead of those of the blktap2 device. This
causes the udev hotplug scripts to write the /physical-device/ entry with
information of the LV, creating the blkfront-blkback-vdi datapath without
blktap2 or tapdisk3 in the middle.

Note that a tapdisk3 and a blktap2 will still exist in dom0, providing all
necessary backend infrastructure resources for other operations.

Signed-off-by: Felipe Franciosi <felipe@paradoxo.org>


chandrikas:
Reviewed and tested. Looks good to merge @franciozzy

franciozzy:
Placed in the PQ.

PR #252, opened 2015-05-19
CP-12403: Set backend-kind to vbd for all VDIs in LUNperVDI SRs

franciozzy:
When tapdisk3 was introduced in Creedence, we needed to ensure that VDIs in
LUNperVDI SRs continued to be plugged via blkback. This was done for the
RawHBA SR (which overloads the _query() method) but not for the ISCSI SR.

This patch does it on the LUNperVDI class, fixing it for the ISCSI SR.

The fix consists of adding an sm-config key to all VDIs in such an SR
instructing the toolstack to write in a .../vbd/... path instead of the
default .../vbd3/...

Signed-off-by: Felipe Franciosi <felipe@paradoxo.org>


chandrikas:
Looks good and works. Please merge.

PR #254, opened 2015-05-27
CA-165611: DM Multipath mishandles blacklisted devices if some paths down

kostaslamda:
This bug may occur when [1 <= paths < total_num_paths] to a device are offline
and the device is blacklisted. Up until now, we checked only the first path in
the list, which might be offline. If this was the case, we incorrectly
determined that the device was not blacklisted (the lack of the
"wwid blacklisted" string doesn't mean it's whitelisted).

To tackle the issue, we check all paths to the device until we have a definite
answer.

Signed-off-by: Kostas Ladopoulos <konstantinos.ladopoulos@citrix.com>


germanop:
Reviewed-by: Germano Percossi <germano.percossi@citrix.com>

PR #256, opened 2015-05-27
CA-171849: Symlink in /dev/disk/by-scsid failed for dm devices

germanop:
XS ships with its own udev rules for multipath and scsi devices.
It is possible that udev's behaviour changed but the net result now
is that we are unable to create a symlink for dm devices in /dev/disk/by-scsid.

/dev/disk/by-scsid is a custom dir we for standard scsi devices and it is
correctly populated but when it comes to dm devices built on top of those
(as it is the case for multipath devices), the link is not there.

The reason being, dm creation is always a 2 steps process
(from udev point of view):
the device is added (action = add) empty and then its status changes
(action = change) to make it point to the scsi devices to use for multipathing.

The rule contained in sm creates the link only during add.
In fact, the symlink is there after addition (you can test this using
"udevadm test") but when the change action kicks in, the link is removed
due to some internal udev rules about symlink management.

The fix is just to create the symlink regardless of the action
(when the device is removed udev is clever enough to remove the link itself).

Easy way to test it:

* create multiple connections to the same device (e.g. different paths to same iscsi target)
* check those device are listed under /dev/disk/by-scsid/<scsid>/ (they are)
* create a multipath device on top of them
* check a corresponding m-* device is created
* check there is now a "mapper" entry in the above dir

The last step fails without the fix.

Signed-off-by: Germano Percossi <germano.percossi@citrix.com>


germanop:
I understand you concern but this is not the case.
Even though there are race conditions that udev is not able to deal with, in this specific case
we are running a program to get the scsid if the device so, if the device is going to be removed
and another one added, if the program is run while the device has been removed and the new one
did not come up, it will fail; if the command is run on the new device, the scsid will be different, hence it will into the right dir, given the dir is named  after the scsid.

I think that however, that udev is clever enough to do the right thing in case of remove (it does not create anything pointing to a device to be remove because it knows it is going to be remove).

That's said, for the sake of clarity, I will make explicit what it is meant here.
Moreover, I do not like crappy rules, so I am gonna make this a little bit better even though
the rest of our custom rules need some "polishing".

Fixed the typos.

kostaslamda:
Repro steps followed; pull request addresses the problem.

germanop:
This fix is interfering with thin provisioning. Hold until investigated further

PR #258, opened 2015-06-08
CA-155766 Fix race between lvchange and tap-ctl open

stefanopanella:
The new lvm2 code shipped with CentOS 7 has a different
behaviour from the previous one since it de-activates
a logical volume while changing permission from "rw" to
"r". This change implies that if 'lvchange -p r' is executed
whlile a tap-ctl open is happening a volume in the chain can
just disappear and the open fails with the error mentioned
in CA-155766.

I have enforced:

- a lv lock for every lvchange -p
- a lock for every lv in the chain while executing
  tap-ctl open

Signed-off-by: Stefano Panella <stefano.panella@citrix.com>

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -41,6 +41,7 @@ from SR import SROSError
 
 import resetvdis
 import vhdutil
+import lvhdutil
 
 # For RRDD Plugin Registration
 from SocketServer import UnixStreamServer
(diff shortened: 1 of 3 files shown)


Zaitypola:
Code review was successful. It is ready to be shipped.

- Jorge Calvo Martin <jorgema@citrix.com>

stefanopanella:
I have modified the pull request since It was breaking nfs SRs.
I have added extra checks befor locking/unlocking the parent chain of the VHD we want to "tap-ctl open"

martinjorge:
Tests with NFSSR are now passing after latest changes. The code is ready.

PR #259, opened 2015-06-09
CP-9913: Added originator for xapi login_with_password

koushikcgit:
The login_with_password XenAPI call takes an "originator"
string as its fourth parameter.

If a client does this, then it gets its own pool of xapi sessions.
Moreover it will not have its xapi sessions destroyed prematurely
as a result of some other misbehaving client that keeps creating
sessions and not logging out of them.

This patch adds the "originator" to all invokes of
login_with_password.

Signed-off-by: Koushik Chakravarty <koushik.chakravarty@citrix.com>

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -1006,7 +1006,7 @@ class VDI(object):
         import XenAPI
 
         session = XenAPI.xapi_local()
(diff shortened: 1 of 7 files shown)


germanop:
Acked-by: Germano Percossi <germano.percossi@citrix.com>

PR #264, opened 2015-07-07
CA-174104: Setting LVM_SYSTEM_DIR to the correct value for trim

pritha-srivastava:
Commands like lvcreate and lvremove invoked as part of trim do
not work, since LVM_SYSTEM_DIR is not being set to the correct
value in the code. Setting this environment variable to the
correct value, allows modification of vg metadata.

Signed-off-by: Pritha Srivastava <pritha.srivastava@citrix.com>


chandrikas:
Looks good and works. Ready to be accepted.

PR #266, opened 2015-07-08
CA-174892: scsiutil.getSCSIid() fails if multipath device is given

kostaslamda:
When getSCSIid() is passed a multipath device (e.g. /dev/dm-2) it fails,
as rawdev() strips any numbers in the end of the device name. Firstly,
rawdev() is modified to return device names that contain 'dm-' untouched.
Secondly, getSCSIid() uses 'path' with the new version of the command,
and only searches for the raw device name in the fallback call.

Signed-off-by: Kostas Ladopoulos <Konstantinos.Ladopoulos@citrix.com>

### DIFF
--- a/drivers/scsiutil.py
+++ b/drivers/scsiutil.py
@@ -31,6 +31,7 @@ import mpath_cli
 PREFIX_LEN = 4
 SUFFIX_LEN = 12
 SECTOR_SHIFT = 9
+SCSI_ID_BIN = '/usr/lib/udev/scsi_id'
 
 def gen_hash(st, len):
     hs = 0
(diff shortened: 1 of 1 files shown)


kostaslamda:
getdev() resolves all symlinks and returns the block device's name. In the case of multipath devices I believe it's always 'dm-#'.

franciozzy:
That doesn't mean other block devices couldn't contain the substring "dm-" in their names. These would mistakenly be taken as device mappers, whereas comparing the major number is foolproof.

germanop:
Felipe, this is not a way to test if a device is a device mapper device, this is just to avoid stripping numbers for devices like dm-0, where the number in this case is not a partition.

Calling a device dm-* is a device mapper thing, according to udev rules.
If another device type is mistakenly using this number then anything could happen.
They could even use their major numbers if they are crazy enough.

germanop:
@kostaslamda, I still think that checking for the name is fine but I agree with Felipe that searching for a substring "dm-" is not sufficient. "starting with" is a stronger one.

germanop:
OK

germanop:
Kostas, I am not entirely sure but I think this fix is superseded by #269

kostaslamda:
At a first glance, it seems that #269 has the exact opposite functionality; you supply a SCSI id and you get back all the block devices associated with that id.

germanop:
Nope, it is exactly for the same thing: it makes sure a device mapper is never passed, making your fix moot.

kostaslamda:
I disaggree. You can remove if you want, but it will just break stuff, without providing any improvement whatsoever.

germanop:
If you disagree just explain why.

kostaslamda:
getSCSIid() takes a block device path as input and returns its SCSI id. Different paths can resolve to the same SCSI id (partitions, multipathed block device, symlinks). getSCSIid() queries a block device about its SCSI id. get_devices_by_SCSIid() takes a SCSI id as input and returns a list of block devices. #269 is an 'ls' on '/dev/disk/by-scsid/<SCSIid>', removing the 'mapper' entry if it exists.

germanop:
Thanks for the explanation about what the functions do, even though I am pretty sure I know both the functions very well..

I was just asking you to verify that that fix did not have any impact on your patch, for example,
preventing getSCSIid from getting into a situation where it get a dm-* device.
It is clear to me the functions do not do the same thing.

What was not clear if the fix in #269 was somehow improving the situation there, because it
explicitly remove a dm device from the list.

While this conversation was taking place I checked myself if that was the case.
Never mind

PR #268, opened 2015-07-17
CA-176801: Work around rpcinfo intricacies to determine nfs versions

siddharthv:

### DIFF
--- a/drivers/nfs.py
+++ b/drivers/nfs.py
@@ -65,9 +65,8 @@ def check_server_tcp(server, nfsversion=DEFAULT_NFSVERSION):
     False otherwise.
     """
     try:
-        util.ioretry(lambda: util.pread([RPCINFO_BIN, "-t",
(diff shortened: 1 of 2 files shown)


sriccio:
While changes are being made to nfs.py, would it be possible to allow to specify mount options for NFS storage repositories. Like not everybody wants to use soft mount, to avoid virtual machines switching to read only when a disconnection between the hosts and the storage occurs.

Actually XenCenter allowed or still allows, I've not really checked recently, to specify mount options, when creating a NFS repository. But nothing in nfs.py is caring about these options :/

Your thoughts ?

siddharthv:
@sriccio thanks a lot for your comment.
I can see what you're faced with but it would be best if you could raise a `Feature Request` at `http://bugs.xenserver.org/` to get such a change onto the product. 
This change is not trivial in my opinion.

PR #271, opened 2015-08-27
CA-180792: Enable ALUA autodetection and group_by_prio for VPLEX

robertbreker:
In CA-141002 we have enabled ALUA autodetection for VNX. This commit enables
the same capability as well for VPLEX.
In addition to enable the autodetection, this commit also changes the
path_grouping_policy to group_by_prio, as the current upstream setting
multibus wouldn't make sense for ALUA.

The partner has completed the qualification of the updated multipath
configuration with their storage arrays. They have tested both, ALUA and
legacy mode.

Signed-off-by: Robert Breker <robert.breker@citrix.com>

### DIFF
--- a/multipath/multipath.conf
+++ b/multipath/multipath.conf
@@ -42,6 +42,13 @@ devices {
 		retain_attached_hw_handler yes
 	}
 	device {
+		vendor			"EMC"
+		product			"Invista"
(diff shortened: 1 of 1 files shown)


germanop:

PR #272, opened 2015-09-03
CA-181447: Fix checking of node minor

germanop:
stat --format=%T returns minor device type in hex and this value is compared
against ROOT_DISK_MINOR which is decimal. Convert value returned by stat %T
to decimal, before doing the comparison.

Signed-off-by: Anoob Soman <anoob.soman@citrix.com>
Reviewed-by: Germano Percossi <germano.percossi@citrix.com>


germanop:
OK

PR #274, opened 2015-09-09
CA-136041: Multipathd forgets faulty paths after an SR operation

geosharath:
The operations pbd-plug and -unplug cause 'multipathd -k "reconfigure"'.
During reconfigure all(!) maps are deleted and re-created. Multipathd
re-creates maps without paths that are temporarily down when doing this.
These paths will only restore if they timeout beyond their dev_loss_tmo
and get re-added later, as this triggers the multipathd behaviour for new
devices. Hence, any SR operation done before dev_loss_tmo the faulty paths
are lost and never added. The fix to this problem is two fold -

a) multipathd reconfigure has been modified to include even faulty paths
 while recreating maps

b) devices have to be deleted from system when SR operations are performed
 before dev_loss_tmo

### DIFF
--- a/drivers/mpath_dmp.py
+++ b/drivers/mpath_dmp.py
@@ -122,14 +122,22 @@ def _resetDMP(sid,explicit_unmap=False,delete_nodes=False):
         except:
             util.SMlog("WARNING: exception raised while attempting to"
                        " modify multipath.conf")
-        try:
+
+# Overriding delete_nodes value to True, since reconfigure does not drop
+# devices that are faulty during pbd unplug and plug. They remain 
+# the system which prevents creation of new maps. This setting is valid only 
+# for any SR operation done before dev_loss_tmo, post which the device is
+# automatically dropped by the kernal.
+        
+	delete_nodes=True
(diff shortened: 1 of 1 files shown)


germanop:
I added a sleep after deleting the nodes to give time to the system to settle.
In fact, after sr-create, sr_attach is issued immediately by xapi and, in some cases, it finds devices
that are in the middle of a remove operation but still visible.
Because of that, we think we do not need to rescan the bus, finding ourselves without valid devices for multipath few lines later

germanop:
@geosharath, I am going to close this PR for the following reasons:
* Device removal could not be as quick as expected by later calls, as explained above
* Removing the devices explicitly is something that goes against vendor specific settings specified
in multipath configuration (on one hand we allow them to specify their favourite settings and on the other hand we device removal)
* Rescanning the bus is too expensive to be done without a valid reason
* The counterpart of this patch in multipath is dangerous, because we do not let "reconfigure" undo
the paths that are down but if those paths are in the system while doing the mapping, the mapping
will fail. This is a corner case less likely than the one we are trying to address but still likely.
* We do have an alternative approach

PR #275, opened 2015-09-09
CA-125111: mpathutil does not work with latest multipathd

geosharath:
Fix: According to the documentation end uses are allowed to use only
mpathutil list and mpathutil status. They are refrained from using
mpathutil add, mpathutil remove and mpathutil resize, hence removing
these features from mpathutil.

Signed-off-by: Sharath Babu sharath.babu@citrix.com

### DIFF
--- a/drivers/mpathutil.py
+++ b/drivers/mpathutil.py
@@ -21,18 +21,6 @@ import scsiutil
 import os
 import time
 
-def add(scsi_id):
-    devices = scsiutil._genReverseSCSIidmap(scsi_id)
-    for device in devices:
-        realpath = os.path.realpath(device)
-        base = os.path.basename(realpath)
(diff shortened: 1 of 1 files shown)


stefanopanella:
After all discussion happened we have agreed that it is fine to just leave list and status as it is mentioned the documentation.
This means the patch can go in as it is.

stefanopanella:
This has been added to sm.pg (5931a938381fc3819f79b5a9d4bcb60d7210f44f)

germanop:
Reviewed-by: Stefano Panella <stefano.panella@citrix.com>

germanop:
The original patch
https://patch-diff.githubusercontent.com/raw/xapi-project/sm/pull/275.patch
contained root@eddie as author and this is what I blindly pushed

PR #276, opened 2015-09-21
CA-182713  : lvm.conf file is wrong in some systems

geosharath:
This happens only when we upgrade SM rpm. During an upgrade, after installation
    of the new package, old package is removed and %postun is called. In %postun
    all changes done to the lvm.conf file gets reset to lvm.conf.orig which is the
    expected behavior during package uninstall and not during package upgrade. As
    a fix to this issue a logic is added in %postun to differentiate between package
    uninstall and package upgrade. This prevents lvm.conf to be reset during a
    package upgrade.

### DIFF
--- a/mk/sm.spec.in
+++ b/mk/sm.spec.in
@@ -61,8 +61,12 @@ exit 0
 
 %postun
 %systemd_postun make-dummy-sr.service
-[ ! -d /etc/lvm/master ] || rm -Rf /etc/lvm/master || exit $?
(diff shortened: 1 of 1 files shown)


stefanopanella:
This has been added to sm.pg (5931a938381fc3819f79b5a9d4bcb60d7210f44f)

germanop:
Reviewed-by: Stefano Panella <stefano.panella@citrix.com>

PR #279, opened 2015-09-30
CA-184022 : Exception happening in _deactivate_locked cause

geosharath:
backend link not to be removed

This exception occus when the network buffer gets filled. Python
server reads the data from the buffer in chunks and not as a whole.
Hence, during heavy traffic conditions the server is forced to advertise
the receiving window size as '0' multiple times which terminates the
connection at the remote end.

As a fix we create a session inside a loop with a fixed amount of tries,
this will introduce sufficient delay for the buffer to get cleared and a
session to be created. Keeping in mind scalability we can additionally introduce
a short delay. This error is bound to occur across XenServer where ever we are
trying to create a session, hence it would be good to foresee this issue and handle
it everytime we create a new session and the same has been done.

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -1960,14 +1960,10 @@ class VDI(object):
             util.SMlog("ERROR: Local cache SR not specified, ignore")
             return
 
-        session = XenAPI.xapi_local()
-        session.xenapi.login_with_password('root', '', '', 'SM')
-
         if caching:
-            self._remove_cache(session, local_sr_uuid)
+            self._remove_cache(self._session, local_sr_uuid)
 
-        self._updateCacheRecord(session, self.target.vdi.uuid, None, None)
-        session.xenapi.session.logout()
+        self._updateCacheRecord(self._session, self.target.vdi.uuid, None, None)
 
     def _is_tapdisk_in_use(self, minor):
         (retVal, links) = util.findRunningProcessOrOpenFile("tapdisk")


chandrikas:
@geosharath, I think the solution has to be two-fold

a) Fix the root cause of the exception. Your solution will help here even though I am not really in favour of touching so many files as part of this bugfix. Also, there are missing imports (time etc.) in a lot of files. 
b) Error handling inside _deactivate. Your PR doesn't address this problem and I think this is the main issue. We might encounter an exception anywhere inside _deactivate and should handle that gracefully. Is that possible?

geosharath:
@chandrikas  Thanks for pointing out the missing import files, I will add them. 
Regarding your second point, this is what I had in mind while coming up with this fix, please correct me if am wrong. You use exception handling when you know things might fail, however in our case _deactivate must not fail, it has do be atomic in nature. If it fails half way then it leaves the system in an inconsistent state which cannot be undone. Hence, I primarily concentrated in preventing the _deactivate from failing rather than handling the scenario where it fails because it anyways doesn't help. Incase if this has to be handled then it can be done while calling _deactivate. Please share your thoughts.

chandrikas:
Let me rephrase that - it was more of a question :-) Can we handle exceptions gracefully in _deactivate and roll back? I guess the answer is no. It's not a good answer but that's the answer we have right now. 

As for your fix, let's see if we can simplify it. Historically, we have been a little lazy about reusing xapi sessions. We just create a new session when needed. Can we check if self._session can be reused for this? I don't have a host to test this right now but that would be ideal.

geosharath:
@chandrikas I have made the changes according to your comments, testing self._session took time since I have to create a setup which naturally follows the code path (intelli cache) and triggers this function. Yes, to be honest its not a good answer that we cannot handle the exception gracefully. This step is the last step while we shutdown a VM, the tapdisk has already been destroyed, we can roll back the changes as you suggested by creating a new tapdisk. Then the minor number of the new tapdisk will definitely vary and XAPI will still remember the old minor number  which will still lead to the issue in CA-184022.

PR #280, opened 2015-10-05
CA-174964: VM.start failed, global name 'XenAPI' is not defined

geosharath:
This problem occurs with VMs that went down with physical host suddenly
stopping(crash/power outage etc). If such VM was exported and then this
pool had host removed post exporting a VM, or if VM is imported to another
pool and it still has host_OpaqueRef in the VDI sm_record, this VM will
fail to import

 Mark Starikov identified this issue and initially proposed a solution that
 removes host_OpaqueRef entry while exporting, however we tend to try not to
 fiddle with the internal data of the SM backends as much as possible. Hence,
 as a fix to this issue a logic is added during reset of a VDI which handles
 host_OpaqueRef that is not present in the database.

### DIFF
--- a/drivers/resetvdis.py
+++ b/drivers/resetvdis.py
@@ -20,6 +20,7 @@
 
 import util
 import lock
+import XenAPI
 
 def reset_sr(session, host_uuid, sr_uuid, is_sr_master):
     from vhdutil import LOCK_TYPE_SR
@@ -62,6 +63,7 @@ def reset_vdi(session, vdi_uuid, force, term_output=True, writable=True):
         if key.startswith("host_"):
             host_ref = key[len("host_"):]
             host_uuid = None
+            host_invalid = False
             host_str = host_ref
             try:
                 host_rec = session.xenapi.host.get_record(host_ref)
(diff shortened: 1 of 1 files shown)


germanop:
Code is functionally OK.

There should be in the commit message, in my opinion, a credit to Mark to have found the problem
and provided an initial patch.
Morevoer, remove all the tabs used in place of spaces (there 3 lines starting with tabs).

I was going to argue about another couple of things but I noticed that you just stuck with the surrounding code.
I am writing that just as an optional thing and for future reference:
* we should be using the more sophisticated .format function instead of % for strings; it was not used in the past just because Python did not have it but it is better to leave it as it is
* we should avoid breaking lines using \ as much as possible (the previous point would have helped even in this case
* I am a bit worried about the lack of "try" around the invocations of xapi commands: they can fail and we won't do anything about that but, again, I see that the surrounding code does (wrongly) not have this mechanism as well.

geosharath:
@germanop I have made the necessary changes, please let me know if you have more concerns.

germanop:
Reviewed-by: Germano Percossi \<germano.percossi@citrix.com\>

PR #282, opened 2015-10-14
CA-184394: Stress Test failure: Job 1186538

geosharath:
Added a logic to check if a dm device exists before attempting to
remove it

### DIFF
--- a/drivers/lvutil.py
+++ b/drivers/lvutil.py
@@ -972,7 +972,15 @@ def _lvmBugCleanup(path):
             except util.CommandException, e:
                 if i < LVM_FAIL_RETRIES - 1:
                     util.SMlog("Failed on try %d, retrying" % i)
-                    time.sleep(1)
+                    cmd = [CMD_DMSETUP, "status", mapperDevice]
+                    try:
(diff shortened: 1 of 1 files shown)


drivers/lvutil.py
```diff
@@ -972,7 +972,13 @@ def _lvmBugCleanup(path):
             except util.CommandException, e:
                 if i < LVM_FAIL_RETRIES - 1:
                     util.SMlog("Failed on try %d, retrying" % i)
-                    time.sleep(1)
+                    cmd = [CMD_DMSETUP, "status", mapperDevice]
+                    try:
+                        util.pread(cmd, expect_rc=1)
+                        util.SMlog("_lvmBugCleanup: dm device %s removed" % mapperDevice)
```
germanop: Line too long.
Please, use format() for formatting strings instead of %s

Use any pep8 compliant program to check syntax (I suggest pep8radius).

germanop:
The patch is functionally OK but please fix the line length and make sure we use format().

Fix the first line of the commit message. They should reflect what the patch is doing not
the reported automated test failure

geosharath:
Sure, will incorporate these changes.

PR #295, opened 2016-01-28
Changes for CA-196882, FileSR and GC changes

MarkSymsCtx:
GC in batch mode and deletes done immediately in SRs
FileSR performs VDI.delete inline
Inline delete VDIs for LVHD SRs
GC waits five minutes after start before initiating work loop.
Use two locks in gcLoop, one to prevent multiple GC processs starting
and one to interlink with abort requests while work is actually occuring.
Inline delete orphaned nodes in coalesce
After handling garbage in the GC loop proceed to coalesce without returning
to scan and potentially finding more garbage whcih then prevents coalesce.

Fix bug in _kickGC where the running GC could be stopped but not start
a new one.

### DIFF
--- a/drivers/LVHDSR.py
+++ b/drivers/LVHDSR.py
@@ -1365,8 +1365,6 @@ class LVHDSR(SR.SR):
                     if e.code != errno.ETIMEDOUT:
                         raise
                     util.SMlog('failed to abort the GC')
-                finally:
-                    return
             else:
                 util.SMlog("A GC instance already running, not kicking")
                 return
@@ -1805,6 +1803,13 @@ class LVHDVDI(VDI.VDI):
         # step: GC might have removed the LV by then
         if self.sr.lvActivator.get(self.uuid, False):
             self.sr.lvActivator.deactivate(self.uuid, False)
+
+        try:
+            self.sr.lvmCache.remove(self.lvname)
+        except SR.SRException, e:
+            util.SMlog(
+                "Failed to remove the volume (maybe is leaf coalescing) "
+                "for %s err:%d" % (self.uuid, e.errno))
 
         self.sr._updateStats(self.sr.uuid, -self.size)
         self.sr._kickGC()
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -62,6 +62,11 @@ FLAG_TYPE_ABORT = "abort"     # flag to request aborting of GC/coalesce
 LOCK_TYPE_RUNNING = "running" 
 lockRunning = None
 
+# process "lock" to indicate that the GC process has been activated but may not
+# yet be running, stops a second process from being started.
+LOCK_TYPE_GC_ACTIVE = "gc_active"
+lockActive = None
+
 # Default coalesce error rate limit, in messages per minute. A zero value
 # disables throttling, and a negative value disables error reporting.
 DEFAULT_COALESCE_ERR_RATE = 1.0/60
@@ -1456,6 +1461,7 @@ class SR:
         for vdi in self.vdis.values():
             if vdi.isCoalesceable() and vdi not in self._failedCoalesceTargets:
                 candidates.append(vdi)
+                Util.log("%s is coalescable" % vdi.uuid)
 
         # pick one in the tallest tree
         treeHeight = dict()
@@ -1748,6 +1754,7 @@ class SR:
 
         vdi.parent._reloadChildren(vdi)
         self.journaler.remove(vdi.JRN_RELINK, vdi.uuid)
+        self.deleteVDI(vdi)
 
     def _coalesceLeaf(self, vdi):
         """Leaf-coalesce VDI vdi. Return true if we succeed, false if we cannot
@@ -1786,7 +1793,6 @@ class SR:
         Util.log("Coalescing parent %s" % tempSnap)
         util.fistpoint.activate("LVHDRT_coaleaf_delay_2", self.uuid)
         self._coalesce(tempSnap)
-        self.deleteVDI(tempSnap)
         if not vdi.isLeafCoalesceable():
             Util.log("The VDI tree appears to have been altered since")
             return False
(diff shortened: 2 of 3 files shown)


drivers/cleanup.py
```diff
@@ -2752,7 +2771,7 @@ def should_preempt(session, srUuid):
     if len(entries) == 0:
         return False
     elif len(entries) > 1:
-        raise util.SMException("More than one coalesce entry: " + entries)
+        raise util.SMException("More than one coalesce entry: " + len(entries))
```
germanop: I think the purpose of the old code was to list the entries, not just report the number

drivers/cleanup.py
```diff
@@ -2532,48 +2538,58 @@ def normalizeType(type):
...
-                sr.xapi.srUpdate()
-                continue
+            if not lockRunning.acquireNoblock():
+                Util.log("Unable to acquire GC running lock.")
+                return
+            try:
+                if not sr.gcEnabled():
+                    break
+                sr.cleanupCoalesceJournals()
+                sr.scanLocked()
```
germanop: I thought we were going to get rid of one of the calls to scanLocked

MarkSymsCtx: It's not clear that sr.cleanupCoalesceJournals doesn't require the scan to be repeated. It is a possible future improvement if this is not the case.

drivers/cleanup.py
MarkSymsCtx: Update to replace len with str.

drivers/LVHDSR.py
```diff
@@ -1806,6 +1804,11 @@ def delete(self, sr_uuid, vdi_uuid):
         if self.sr.lvActivator.get(self.uuid, False):
             self.sr.lvActivator.deactivate(self.uuid, False)
 
+        try:
+            self.sr.lvmCache.remove(self.lvname)
+        except e:
+            util.SMlog("Failed to remove the volume (maybe is leaf coalescing) for %s err:%d" % (self.uuid, e.errno))
```
germanop: Too long, suggested
```
            util.SMlog(
                "Failed to remove the volume (maybe is leaf coalescing) for %s err:%d" %
                (self.uuid, e.errno))
```

drivers/cleanup.py
```diff
@@ -2532,48 +2538,58 @@ def normalizeType(type):
...
+
+                howmany = len(sr.findGarbage())
+                if howmany > 0:
+                    Util.log("Found %d orphaned vdis" % howmany)
+                    sr.garbageCollect(dryRun)
+                    sr.xapi.srUpdate()
+
+                candidate = sr.findCoalesceable()
+                if candidate:
+                    util.fistpoint.activate("LVHDRT_finding_a_suitable_pair",sr.uuid)
```
germanop: I know it is not part of your code but it is a part we already touching, so pep8 is suggesting
```
                    util.fistpoint.activate(
                        "LVHDRT_finding_a_suitable_pair", sr.uuid)
```

drivers/LVHDSR.py
MarkSymsCtx: Also too long, this fits into 80 chars

`            util.SMlog(
                "Failed to remove the volume (maybe is leaf coalescing) "
                "for %s err:%d" % (self.uuid, e.errno))
`

chandrikas:
New change has been added to patchqueue

drivers/cleanup.py
```diff
@@ -2532,48 +2538,58 @@ def normalizeType(type):
...
+                sr.cleanupCoalesceJournals()
+                sr.scanLocked()
+                sr.updateBlockInfo()
+
+                howmany = len(sr.findGarbage())
+                if howmany > 0:
+                    Util.log("Found %d orphaned vdis" % howmany)
+                    sr.garbageCollect(dryRun)
+                    sr.xapi.srUpdate()
+
```
germanop: What is the reason for "continue" to disappear from here?

MarkSymsCtx: It's not required and allows the coalesce to start without continually finding garbage. As it happens with the inlining of the deletes in the SR and the coalesce the garbage path is never taken (unless in case of crash during LV delete).

drivers/LVHDSR.py
```diff
@@ -1365,8 +1365,6 @@ def _kickGC(self):
                     if e.code != errno.ETIMEDOUT:
                         raise
                     util.SMlog('failed to abort the GC')
-                finally:
```
germanop: Why are we dropping this?

germanop:
Apart from this, code is fine

drivers/LVHDSR.py
MarkSymsCtx: The 

```
finally: 
    return

```
caused the running GC to be stopped but didn't start a new one. There probably should be a return in the case of an exception however

drivers/cleanup.py
germanop: Not addressed

drivers/cleanup.py
```diff
@@ -2741,8 +2760,8 @@ def get_state(srUuid):
     locking.
     """
     init(srUuid)
-    if lockRunning.acquireNoblock():
-        lockRunning.release()
+    if lockActive.acquireNoblock():
```
germanop: This change was not here in the previous version.
What do we want to tell the user: if it is spawned or actively running?

MarkSymsCtx: Spawned, this was in the patch queue but missed from the push. This is what caused the BVT to fail in the first submission.

germanop:
About the commit message: I appreciate it is the squash of the two previous commits but the
first line should convey both info not just the one from the first commit squashed.

Let's find a one line title that captures the nature of the fix.
If there is one CA ticket more descriptive than the other, use that only but the first
line should describe the change in its entirety.

Something like "GC in batch mode and deletes done immediately in SRs", for example.
You know better.

PR #299, opened 2016-02-04
CP-6096: Extract out raw iscsi SR

BobBall:
The existing 'iscsi' SR type functionally works very well, however ISCSISR a base class for LVHDoISCSI and OCFSoISCSI SRs, therefore it would be risky to claim support for raw ISCSI connections in OpenStack, since some issues may not be fixable without causing complications for the existing supported SR types.

This patch series creates a new leaf SR type for 'iscsi' by moving the existing SR type to a new base class


BobBall:
Also verified through XenRT job 1300229

germanop:
In your second commit you did rename ISCSISR.py to ISCSI_base.py but, in the same commit, you immediately created a new, almost empty, ISCSISR.py

That does not work well when git tries to understand renames..

germanop:
I have checked the second commit: ISCSISR.py became a symlink to ISCSI_base. Do not think it is necessary, especially because in the last commit you replace it with a new 
regular file

germanop:
Regardless of my raised points, your second commit does not build because you use
BaseISCSISR but the module does not have it..

germanop:
* It is fine if you call the module itself BaseISCSI instead of ISCSI_base, so we won't break are custom
to have the SR class named after the module (in the last commit, so the second will show up as a rename only)
* It is fine if we maintain the SR type as "iscsi" and for the symlink "ISCSISR" but we really need the name of the python file to be changed into something like "RawISCSISR.py" otherwise other component can still used the old name silently invoking the new module.
* In fact, in xencert we import ISCSISR (that needs to be changed, too)

germanop:
@geosharath @chandrikas could you comment on the impact of this change on xencert?

germanop:
HBASR.py is still import from ISCSISR

germanop:
@BobBall what is your intention for test_ISCSISR.py? Who needs to be tested, ISCSISR or ISCSI_base. In the latter case it is importing ISCSISR

BobBall:
I'd argue that test_ISCSISR.py should be testing the SR type itself, not the base class.

tests/test_ISCSISR.py
```diff
@@ -88,8 +88,8 @@ def _init_adapters(self):
 
 class TestVdiTypeSetting(TestBase):
 
-    @mock.patch('ISCSISR.iscsilib.discovery')
-    @mock.patch('ISCSISR.iscsilib.ensure_daemon_running_ok')
+    @mock.patch('ISCSISR.ISCSI_base.iscsilib.discovery')
```
germanop: I could be wrong but these changes (the whole series) do not quite seem OK.

germanop:
Here you can find a preliminary version with minor fixes https://github.com/germanop/sm/tree/CP-6069
* Fixed the second commit where BaseISCSISR was used but did not exist
* Removed the symlink ISCSISR.py (so in the second commit this file name does not exist at all)
* Removed the import in HBASR.py referring to ISCSISR (removed not renamed because it is not used)

As a note for myself, pylint seems to be "broken" given I built a package while ISCSISR.py did not exist and pylint accepted HBASR.py that was importing it.
HBASR.py and, in turn, RawHBA.py were silently broken (our tests did not pick it up because the third commit reinstated the file but this is would have been an issue if we just wanted to rename the file, as we should do..)

Remaining items:
- [ ]  I am not sure the tests should test ISCSISR.py (or RawISCSISR.py if we rename it)
- [ ] Not sure your modifications to test_ISCSISR.py works
- [ ] Rename ISCSISR.py to RawISCSISR.py (and hence, everything referring to it) to avoid other components silently importing the wrong module instead of failing because the module is not found
- [ ] Rename ISCSI_base to BaseISCSI (or change the name of its class accordingly)

BobBall:
Verified through XenRT job 1370416 with a seperate SM RPM

germanop:
Reviewed-by: Germano Percossi \<germano.percossi@citrix.com\>

PR #297, opened 2016-02-01
CP-15900: Handle new Tools ISO name

robhoes:
The name of the Tools ISO is going to change from "xs-tools.iso" to
"tools.iso". This patch changes the ISOSR backend to recognise both names, and
set the VDI.name_label accordingly, to simplify the transition.

Signed-off-by: Rob Hoes <rob.hoes@citrix.com>

PR #300, opened 2016-02-10
CP-15900: Handle new Tools ISO name

robhoes:


robhoes:
This is a followup to #297. This PR includes the patch from the other PR, because the latter was not yet merged and the new patch depends on it.

chandrikas:
New changes have been added to patchqueue

germanop:
@robhoes do they need to be 2 commits?

robhoes:
@germanop Nope, it just ended up this way because minds were changed along the way. Shall I squash the patches in the PR and close the other?

germanop:
Close all the PR that are not needed. For squashing, it is not a problem. If it needs to be one commit only, just tell me which commit message has to be kept otherwise I can push both commits.
But, yes, if you want to rearrange the commits in this PR, it is more than welcome.

I will compare it again with what we have in our patch queue to check there are no differences in code.

robhoes:
@germanop I've now combined the two patches into one, and have closed #297.

germanop:
Thank you. I will close this as soon as we can.

PR #301, opened 2016-02-18
CP-15135: Converted updatempppathd init script to systemd unit file

chandrikas:
Converted sm specific init scripts to systemd units. Experimentally removed mpathroot script.


germanop:
Remove the commit for sm-multipath. It is not meant to be a service and it is not called from the 
installer as a service, so this change will break it.

PR #303, opened 2016-03-31
CA-205120: do not start sm-multipath upon RPM installation

germanop:
We need sm-multipath to be started only at boot when we
check if the root device is multipathed or not.

During system installation it throws harmless but
annoying errors.

Signed-off-by: Germano Percossi germano.percossi@citrix.com


germanop:
This change broke the installation for Boot from SAN machines.

It needs either to be improved or rejected and fix the errors thrown differently

PR #304, opened 2016-04-08
Fix LVM unit tests.

BobBall:
The UT depend on a specific path for lvcreate binaries, however
the path actually used is distro dependent.  On Debian distros,
lvcreate is in /sbin.  SM correctly auto detects the location,
but the UT were expecting it in /usr/sbin.

Fix the UT so they use the detected path for the executables.

### DIFF
--- a/tests/lvmlib.py
+++ b/tests/lvmlib.py
@@ -1,4 +1,6 @@
 import optparse
+import lvutil
+import os
 
 
 class LogicalVolume(object):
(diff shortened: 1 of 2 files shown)


germanop:
@MarkSymsCtx this seems reasonable to me but I'd rather have you looking at it

BobBall:
The real issue is that https://github.com/xapi-project/sm/blob/master/drivers/lvutil.py#L40 attempts to detect the location of lv executables on file import.
This means the UT have no real opportunity to provide a stable environment.
Using the detected path seems to be the least-worse change as it doesn't change the functionality.

After approving this change a refactor to fix the lvutil code would at least have UT validating the refactor.

PR #305, opened 2016-04-15
CA-207783: Honour additional options passed to NFS SR creation

chandrikas:

### DIFF
--- a/drivers/ISOSR.py
+++ b/drivers/ISOSR.py
@@ -242,14 +242,15 @@ class ISOSR(SR.SR):
 
         self.credentials = os.path.join("/tmp", util.gen_uuid())
         if self.dconf.has_key('type'):
+            # Creation via XC
             if self.dconf['type']=='cifs':
                 # CIFS specific stuff
                 # Check for username and password
                 mountcmd=["mount.cifs", location, self.mountpoint]
                 self.appendCIFSMountOptions(mountcmd)
         else:
-            # Not-so-legacy mode:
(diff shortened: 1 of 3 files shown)


drivers/ISOSR.py
```diff
@@ -242,13 +242,14 @@ def attach(self, sr_uuid):
 
         self.credentials = os.path.join("/tmp", util.gen_uuid())
         if self.dconf.has_key('type'):
+            # Creation via XC
```
germanop: Why is only XC specific?

chandrikas: That's how we receive it from XAPI. There is an additional type dconf when the request comes in from XC

drivers/NFSSR.py
```diff
@@ -92,6 +92,10 @@ def load(self, sr_uuid):
         if self.dconf.has_key('useUDP') and self.dconf['useUDP'] == 'true':
             self.transport = "udp"
         self.nfsversion = nfs.validate_nfsversion(self.dconf.get('nfsversion'))
+        if self.dconf.has_key('options'):
```
germanop: Not an error and I do understand it is to adhere to code above but
has_key() is deprecated.

Sooner or later we should start "breaking style compatibility" and write standard compliant code.

Up to you

chandrikas: Good point. I'll change it to 
`if 'options' in self.dconf`

germanop:

PR #306, opened 2016-04-19
CA-207781: NFS SR probe should use one of the supported versions

chandrikas:
SR probe for NFS SRs always mounts the temporary share with 
version = 3(default version). If this is not supported by the server, sr-probe
doesn't return any results. Fixed by mounting share with any one of
the valid versions supported by the server.

### DIFF
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -109,7 +109,9 @@ class NFSSR(FileSR.FileSR):
     def check_server(self):
         try:
             if self.dconf.has_key(PROBEVERSION):
-                nfs.check_server_tcp(self.remoteserver)
+                sv = nfs.get_supported_nfs_versions(self.remoteserver)
+                if len(sv):
+                    self.nfsversion = sv[0]
             else:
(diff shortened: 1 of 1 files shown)


drivers/nfs.py
```diff
@@ -65,8 +65,7 @@ def check_server_tcp(server, nfsversion=DEFAULT_NFSVERSION):
     False otherwise.
     """
     try:
-        sv = get_supported_nfs_versions(server)
-        return (True if nfsversion in sv else False)
+        return get_supported_nfs_versions(server)
```
johnmdilley: This appears to change the semantics of this function, which probably means that it no longer does the right thing if called with nfsversion (which it is on NFSSR.py line 116). Although I can't actually see how the return code is used there...

chandrikas: Yes, which is why I added that comment in NFSSR (around line 116). Nobody seems to be doing anything about the return code. As an alternative, I can call get_supported_nfs_versions in NFSSR  (line 112) directly.

johnmdilley:
Looks good

PR #312, opened 2016-05-19
Use a faster check when checking if an ISCSI connection exists

syed:
When we have a lot of ISCSI connections (more than 100) checking for a connection takes a lot of time (order of seconds). This PR does a very simple check to determine if an ISCSI target exisits.

### DIFF
--- a/drivers/iscsilib.py
+++ b/drivers/iscsilib.py
@@ -285,6 +285,7 @@ def login(portal, target, username, password, username_in="", password_in="",
     failuremessage = "Failed to login to target."
     try:
         (stdout,stderr) = exn_on_failure(cmd,failuremessage)
+        wait_for_devs(target, portal)
     except:
         raise xs_errors.XenError('ISCSILogin')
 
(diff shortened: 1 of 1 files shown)


syed:
What do you guys think of this check? It is essentially doing the same thing but I have found it to be way faster.

MarkSymsCtx:
It probably is faster but I don't think it's achieving the same thing. "iscsiadm -m session" will report on the active sessions and therefore verifies that the endpoint is still accessible, just checking /de/iscsi may not be affected by connectivity lose to the iscsi server.

syed:
I see. Thanks Mark. I do have another approach. So the `/dev/iscsi` path looks something like this
```
/dev/iscsi/iqn.2010-01.com.solidfire:giwd.sr-540.3285/172.31.255.160:3260/SESSIONID-98/LUN0
```

Now If we extract the session ID from it (98) and do an `iscsiadm -m seesion -r <session_id?` we get a much faster response. 

```
[root@ucs-xen1 ~]# time iscsiadm -m session -r 98 > /dev/null

real    0m0.054s
user    0m0.000s
sys     0m0.052s
[root@ucs-xen1 ~]# time iscsiadm -m session  > /dev/null

real    0m0.495s
user    0m0.312s
sys     0m0.184s
[root@ucs-xen1 ~]# 
```

What do you think?

MarkSymsCtx:
The second approach sounds reasonable. Please update the pull request.

PR #313, opened 2016-05-19
Don't detach after create

syed:
In `create` we are doing a detach after the ISCSI connection is made which logs us out of the ISCSI session. I am not able to understand the reason behind that as we would eventually attach it anyways.

### DIFF
--- a/drivers/ISCSISR.py
+++ b/drivers/ISCSISR.py
@@ -437,10 +437,7 @@ class ISCSISR(SR.SR):
         if self._loadvdis() > 0:
             scanrecord = SR.ScanRecord(self)
             scanrecord.synchronise()
-        try:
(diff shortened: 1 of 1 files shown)


germanop:
You cannot leave an SR attached after create.
Xapi expects it to be detached, in fact a create operation for xapi is
create, attach for the storage manager

syed:
Ah I see @germanop. Is it possible to d a detach but still keep the ISCSI connection. I say this because I am testing having a large number of ISCSI connections and login/logout for a LUN is very expensive.

germanop:
It is an API choice and it would break many automated tests.
@jonludlam any workaround here?

syed:
Hi, any updates on this guys?

PR #315, opened 2016-06-08
CA-212761: Use eth field in probe data conditionally

chandrikas:
Not all adapters have 'eth' field populated in their probe data.
Check for the existence of the field in the structure before
extracting the value for display.


MarkSymsCtx:
Looks good.

PR #317, opened 2016-06-13
CA-213065: Need to import sys if we read exception data.

MarkSymsCtx:
From: Mark Syms mark.syms@citrix.com


germanop:
@MarkSymsCtx still needed?

MarkSymsCtx:
Yes, needed. It got accidentally closed when I deleted my working branch.

PR #320, opened 2016-07-01
CA-213968: Make NFS retrans configurable

chandrikas:

### DIFF
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -124,7 +124,7 @@ class NFSSR(FileSR.FileSR):
                                      opterr=exc.errstr)
 
 
-    def mount(self, mountpoint, remotepath, timeout = 0):
+    def mount(self, mountpoint, remotepath, timeout = None):
         try:
             nfs.soft_mount(
                     mountpoint, self.remoteserver, remotepath, self.transport,
@@ -138,13 +138,13 @@ class NFSSR(FileSR.FileSR):
         if not self._checkmount():
             self.validate_remotepath(False)
             util._testHost(self.dconf['server'], NFSPORT, 'NFSTarget')
-            #Extract timeout and value, if any
+            #Extract timeout value, if any
             io_timeout = nfs.get_nfs_timeout(self.other_config)
             self.mount_remotepath(sr_uuid, io_timeout)
         self.attached = True
 
 
-    def mount_remotepath(self, sr_uuid, timeout = 0):
+    def mount_remotepath(self, sr_uuid, timeout = None):
         if not self._checkmount():
             # FIXME: What is the purpose of this check_server?
             # It doesn't stop us from continuing if the server
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -79,6 +79,7 @@ class NFSSR(FileSR.FileSR):
         self.nosubdir = False
         if self.sr_ref and self.session is not None :
             self.sm_config = self.session.xenapi.SR.get_sm_config(self.sr_ref)
+            self.other_config = self.session.xenapi.SR.get_other_config(self.sr_ref)
         else:
             self.sm_config = self.srcmd.params.get('sr_sm_config') or {}
         self.nosubdir = self.sm_config.get('nosubdir') == "true"
@@ -137,7 +138,8 @@ class NFSSR(FileSR.FileSR):
         if not self._checkmount():
             self.validate_remotepath(False)
             util._testHost(self.dconf['server'], NFSPORT, 'NFSTarget')
-            io_timeout = nfs.get_nfs_timeout(self.session, sr_uuid)
+            #Extract timeout and value, if any
+            io_timeout = nfs.get_nfs_timeout(self.other_config)
             self.mount_remotepath(sr_uuid, io_timeout)
         self.attached = True
 
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -137,7 +137,7 @@ class NFSSR(FileSR.FileSR):
         if not self._checkmount():
             self.validate_remotepath(False)
             util._testHost(self.dconf['server'], NFSPORT, 'NFSTarget')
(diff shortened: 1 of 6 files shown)


drivers/util.py
```diff
@@ -668,19 +668,22 @@ def get_all_slaves(session):
...
     return filter(lambda x: x != master_ref, host_refs)
 
-def get_nfs_timeout(session, sr_uuid):
+def get_other_config(session, sr_uuid, key):
     if not isinstance(session, XenAPI.Session):
-        SMlog("No XAPI session for getting nfs timeout config")
-        return 0
-
+        SMlog("No XAPI session for getting other config")
+        return None
```
germanop: Why do we return "None" instead of 0?

drivers/util.py
```diff
@@ -668,19 +668,22 @@ def get_all_slaves(session):
     master_ref = get_this_host_ref(session)
     return filter(lambda x: x != master_ref, host_refs)
 
-def get_nfs_timeout(session, sr_uuid):
+def get_other_config(session, sr_uuid, key):
```
germanop: Given this function is not generic, isn't there any way to cache other config to avoid that 2 calls for
different keys will issue 2 calls to xapi?

drivers/util.py
germanop: OK, None is fine in this context if some keys can have 0 as a value

drivers/NFSSR.py
```diff
@@ -137,19 +137,24 @@ def attach(self, sr_uuid):
         if not self._checkmount():
             self.validate_remotepath(False)
             util._testHost(self.dconf['server'], NFSPORT, 'NFSTarget')
-            io_timeout = util.get_nfs_timeout(self.session, sr_uuid)
-            self.mount_remotepath(sr_uuid, io_timeout)
+
+            #Extract timeout and retrans values, if any
+            sr_other_config = util.get_other_config(self.session, sr_uuid)
```
germanop: I meant a property for the SR called other_config that would have been initialized to None in the constructor and
populated the very first time a function needed it.

I think with this patch it is ok to go straight to the point, so keep this and, maybe, we can change things later

drivers/NFSSR.py
```diff
@@ -137,19 +137,24 @@ def attach(self, sr_uuid):
...
+            io_timeout = util.get_nfs_timeout(self.session, sr_uuid, sr_other_config)
+            io_retrans = util.get_nfs_retrans(self.session, sr_uuid, sr_other_config)
+            self.mount_remotepath(sr_uuid, io_timeout, io_retrans)
+            
         self.attached = True
 
 
-    def mount_remotepath(self, sr_uuid, timeout = 0):
+    def mount_remotepath(self, sr_uuid, timeout = 0, retrans = nfs.SOFTMOUNT_RETRANS):
```
germanop: as discussed, let's avoid our own timeout default: stick with mount default if nothing specified

drivers/NFSSR.py
```diff
@@ -137,19 +137,24 @@ def attach(self, sr_uuid):
...
         if not self._checkmount():
             # FIXME: What is the purpose of this check_server?
             # It doesn't stop us from continuing if the server
             # doesn't support the requested version. We fail 
             # in mount instead
             self.check_server()
-            self.mount(self.path, self.remotepath, timeout)
+            self.mount(self.path, self.remotepath, timeout, retrans)
```
germanop: if we do not decide anything about timeout in the calling functions we need default values here as well..
Should we stick with our own default?

drivers/util.py
```diff
@@ -34,6 +34,7 @@
 import traceback
 import glob
 import copy
+import nfs
```
germanop: No more circular deps

germanop:
I think it is time to move the nfs functions out of util.py and make sure no other components are using them

drivers/NFSSR.py
```diff
@@ -137,19 +138,21 @@ def attach(self, sr_uuid):
         if not self._checkmount():
             self.validate_remotepath(False)
             util._testHost(self.dconf['server'], NFSPORT, 'NFSTarget')
-            io_timeout = util.get_nfs_timeout(self.session, sr_uuid)
-            self.mount_remotepath(sr_uuid, io_timeout)
+            #Extract timeout and retrans values, if any
+            io_timeout = nfs.get_nfs_timeout(self.other_config)
+            io_retrans = nfs.get_nfs_retrans(self.other_config)
+            self.mount_remotepath(sr_uuid, io_retrans)
```
germanop: You forgot io_timeout

drivers/blktap2.py
```diff
@@ -1552,7 +1553,9 @@ def activate(self, sr_uuid, vdi_uuid, writable, caching_params):
             # Tapdisk needs the allocation_quantum to be in MB
             options["allocation_quantum"] = int(aq / (1024*1024))
 
-        timeout = util.get_nfs_timeout(self.target.vdi.session, sr_uuid)
+        sr_ref = self.target.vdi.sr.srcmd.params.get('sr_ref')
```
germanop: If we already have a proper SR object maybe we should use the other_config in there, limiting the number of xapi calls. Not sure we have, though.

drivers/blktap2.py
```diff
@@ -1552,7 +1553,9 @@ def activate(self, sr_uuid, vdi_uuid, writable, caching_params):
...
+        sr_config = self._session.xenapi.SR.get_other_config(sr_ref)
```
germanop: sr_other_config, maybe? (and line below, as well)

germanop:
- [ ] forgot io_timeout
- [ ] maybe we can use other_config from cache
- [ ] change of variable name

Apart from this, the commits are split the proper way

chandrikas:
@germanop : I have addressed two out of the three changes requested above. Caching other_config in blktap.py might need a little more work. While it is definitely a "nice-to-have" change, I don't think it is critical at this point.

germanop:
Fine

germanop:
@chandrikas , gentle reminder to update the PR :-)

PR #325, opened 2016-07-29
CA-214746: made snapwatchd service not executable

letsboogey:
daemon.log file complained that snapwatchd.service
file was marked as executable and it said that
executable permission bits had to be removed


MarkSymsCtx:
Change looks good.

PR #327, opened 2016-08-03
CA-207930: Space reclaim operation fails with errors when used with a completely filled SR

swathicitrix:
Signed-off-by: Swathi Kovvuri <swathi.kovvuri@citrix.com>

### DIFF
--- a/drivers/trim_util.py
+++ b/drivers/trim_util.py
@@ -111,15 +111,24 @@ def do_trim(session, args):
 
             # Clean trim LV in case the previous trim attemp failed
             if lvutil.exists(lv_path):
-                lvutil.remove(lv_path)
+               lvutil.remove(lv_path)
 
-            # Perform a lvcreate, blkdiscard and lvremove to trigger trim on the array
-            lvutil.create(lv_name, 0, vg_name, size_in_percentage="100%F")
-            cmd = ["/usr/sbin/blkdiscard", "-v", lv_path]
-            stdout = util.pread2(cmd)
-            util.SMlog("Stdout is %s" % stdout)
-            util.SMlog("Trim on SR: %s complete. " % sr_uuid)
-            result = str(True)
+            #Check if VG limits are enough for creating LV.
+            stats = lvutil._getVGstats(vg_name)
+            if (stats['freespace'] < lvutil.LVM_SIZE_INCREMENT):
+                util.SMlog("No space to claim on a full SR %s" % sr_uuid)
+                err_msg = {ERROR_CODE_KEY: 'Trim failed on full SR',
+                           ERROR_MSG_KEY: 'No space to claim on a full SR'}
+                result = to_xml(err_msg)
+            else:
+                # Perform a lvcreate, blkdiscard and lvremove to
+                # trigger trim on the array
+                lvutil.create(lv_name, 0, vg_name, size_in_percentage="100%F")
+                cmd = ["/usr/sbin/blkdiscard", "-v", lv_path]
+                stdout = util.pread2(cmd)
+                util.SMlog("Stdout is %s" % stdout)
+                util.SMlog("Trim on SR: %s complete. " % sr_uuid)
+                result = str(True)
         except util.CommandException, e:
             err_msg = {
                 ERROR_CODE_KEY: 'TrimException',
--- a/tests/test_trim_util.py
+++ b/tests/test_trim_util.py
@@ -4,6 +4,7 @@ import testlib
 
 import mock
 
+EMPTY_VG_SPACE = 4 * 1024 * 1024
 
 class AlwaysBusyLock(object):
     def acquireNoblock(self):
@@ -99,6 +100,7 @@ class TestTrimUtil(unittest.TestCase, testlib.XmlMixIn):
             sleep.mock_calls
         )
 
+    @mock.patch("trim_util.lvutil.LVM_SIZE_INCREMENT", EMPTY_VG_SPACE)
     @mock.patch('trim_util.lvutil')
     @mock.patch('lock.Lock')
     @mock.patch('util.sr_get_capability')
@@ -108,6 +110,9 @@ class TestTrimUtil(unittest.TestCase, testlib.XmlMixIn):
                                    sr_get_capability,
                                    MockLock,
                                    lvutil):
+        lvutil._getVGstats.return_value = {'physical_size': 0,
+                                           'physical_utilisation': 0,
+                                           'freespace': EMPTY_VG_SPACE}
         MockLock.return_value = AlwaysFreeLock()
         sr_get_capability.return_value = [trim_util.TRIM_CAP]
         context.setup_error_codes()
@@ -119,6 +124,7 @@ class TestTrimUtil(unittest.TestCase, testlib.XmlMixIn):
             size_in_percentage='100%F'
         )
 
+    @mock.patch("trim_util.lvutil.LVM_SIZE_INCREMENT", EMPTY_VG_SPACE)
     @mock.patch('util.pread2')
     @mock.patch('trim_util.lvutil')
     @mock.patch('lock.Lock')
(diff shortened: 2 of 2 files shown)


swathicitrix:
Here All the UT cases related to TRIM functionality got passed.
But below UT is failing , the same error message i could see in failure logs of other commits also.
ERROR: test_load_no_scsiid (test_LVHDoFCoESR.TestFCoESR)

Looks like this is due to travis jobs broken issue.

MarkSymsCtx:
The travis job has been broken since #325. Separate issue to fix.

drivers/lvutil.py
```diff
@@ -391,6 +391,24 @@ def _getLVsize(path):
         raise xs_errors.XenError('VDIUnavailable', \
               opterr='no such VDI %s' % path)
 
+def _getVG_freespace(vgname):
```
chandrikas: Why did we need a new method? Can't _getVGstats be reused?

drivers/trim_util.py
```diff
@@ -104,15 +104,26 @@ def do_trim(session, args):
...
 
     if got_lock:
-        try:
-            vg_name = _vg_by_sr_uuid(sr_uuid)
-            lv_name = sr_uuid + TRIM_LV_TAG
-            lv_path = _lvpath_by_vg_lv_name(vg_name, lv_name)
-
-            # Clean trim LV in case the previous trim attemp failed
-            if lvutil.exists(lv_path):
-                lvutil.remove(lv_path)
```
chandrikas: Why have we moved the following block out of the try block? What if lvutil.remove throws an exception?

drivers/trim_util.py
```diff
@@ -104,15 +104,26 @@ def do_trim(session, args):
...
+        # Clean trim LV in case the previous trim attemp failed
+        if lvutil.exists(lv_path):
+           lvutil.remove(lv_path)
+
+        #Check if VG limits are enough for creating LVM.
+        if lvutil._getVG_freespace(vg_name) < lvutil.LVM_SIZE_INCREMENT:
+           util.SMlog("No space to claim on a completely filled SR %s to perform lvutil.create" % sr_uuid)
+           err_msg = {ERROR_CODE_KEY: 'Trim failed on completely filled SR',
+                      ERROR_MSG_KEY: 'No space to claim on a completely filled SR'}
+           if lvutil.exists(lv_path):
```
chandrikas: This block of code duplicates logic that is already happening further down in do_trim. Can't it be reused?

tests/test_trim_util.py
```diff
@@ -108,6 +110,7 @@ def test_do_trim_creates_an_lv(self,
                                    sr_get_capability,
                                    MockLock,
                                    lvutil):
+        lvutil._getVG_freespace.return_value = EMPTY_VG_SPACE
```
chandrikas: It might be a good idea to mock the freespace in the VG to be significantly more (or at least not equal to) the bare minimum (EMPTY_VG_SPACE)

tests/test_trim_util.py
```diff
@@ -290,3 +299,34 @@ def test_log_last_triggered_exc_logged(self, mock_log_exc, mock_time):
         trim_util._log_last_triggered(session, 'sr_uuid')
 
         self.assertEqual(1, mock_log_exc.call_count)
+
+    @mock.patch("trim_util.lvutil.LVM_SIZE_INCREMENT", EMPTY_VG_SPACE)
+    @mock.patch('trim_util.lvutil')
+    @mock.patch('lock.Lock')
+    @mock.patch('util.sr_get_capability')
+    @testlib.with_context
+    def test_do_trim_creates_an_lv_onfilledSR_not_supported(self,
```
chandrikas: The name of the test case is rather convoluted. Can we have something simple like "test_do_trim_returns_exception_when_sr_full"

drivers/trim_util.py
swathicitrix: Here the reason i moved few lines of code above try block is that, the methods inside
those functions already have exception handling. 

But as you mentioned the reason for these lines inside try block is because we have to give the appropriate TRIM functionality error to end user.

So i moved the code back.

tests/test_trim_util.py
swathicitrix: Can you please let me know the reason here. As it is only for test use, i feel the current
value is sufficient.

drivers/trim_util.py
swathicitrix: Modified the code accordingly.

drivers/trim_util.py
```diff
@@ -111,15 +111,23 @@ def do_trim(session, args):
...
-            # Perform a lvcreate, blkdiscard and lvremove to trigger trim on the array
-            lvutil.create(lv_name, 0, vg_name, size_in_percentage="100%F")
-            cmd = ["/usr/sbin/blkdiscard", "-v", lv_path]
-            stdout = util.pread2(cmd)
-            util.SMlog("Stdout is %s" % stdout)
-            util.SMlog("Trim on SR: %s complete. " % sr_uuid)
-            result = str(True)
+               lvutil.remove(lv_path)
+
+            #Check if VG limits are enough for creating LVM.
```
chandrikas: Change LVM to LV

drivers/trim_util.py
```diff
@@ -111,15 +111,23 @@ def do_trim(session, args):
...
+            stats = lvutil._getVGstats(vg_name)
+            if (stats['freespace'] < lvutil.LVM_SIZE_INCREMENT):
+                util.SMlog("No space to claim on a completely filled SR %s to perform lvutil.create" % sr_uuid)
```
chandrikas: Simplify the comment. "No space to claim on SR to perform lvutil.create" is a little complicated. 
Also, for this and the next couple of lines
- Limit line length to 80 characters. 
- Change "completely filled SR" to "full SR"

tests/test_trim_util.py
```diff
@@ -108,6 +110,7 @@ def test_do_trim_creates_an_lv(self,
                                    sr_get_capability,
                                    MockLock,
                                    lvutil):
+        lvutil._getVGstats.return_value = {'physical_size': 0, 'physical_utilisation': 0, 'freespace': EMPTY_VG_SPACE}
```
chandrikas: Shorten line length here and below

tests/test_trim_util.py
```diff
@@ -290,3 +299,34 @@ def test_log_last_triggered_exc_logged(self, mock_log_exc, mock_time):
...
+                                   MockLock,
+                                   lvutil):
+        lvutil._getVGstats.return_value = {'physical_size': 0, 'physical_utilisation': 0, 'freespace': 0} 
+        MockLock.return_value = AlwaysFreeLock()
+        sr_get_capability.return_value = [trim_util.TRIM_CAP]
+        context.setup_error_codes()
+
+        result = trim_util.do_trim(None, {'sr_uuid': 'some-uuid'})
+
+        self.assertXML("""
```
chandrikas: Change error message to read "full SR" instead of "completely filled SR"

drivers/trim_util.py
swathicitrix: Modified.

drivers/trim_util.py
swathicitrix: Modified in all the places as per this guideline.

drivers/trim_util.py
```diff
@@ -111,15 +111,25 @@ def do_trim(session, args):
...
-            util.SMlog("Stdout is %s" % stdout)
-            util.SMlog("Trim on SR: %s complete. " % sr_uuid)
-            result = str(True)
+               lvutil.remove(lv_path)
+
+            #Check if VG limits are enough for creating LV.
+            stats = lvutil._getVGstats(vg_name)
+            if (stats['freespace'] < lvutil.LVM_SIZE_INCREMENT):
+                util.SMlog("No space to claim on a full SR %s to perform"\
+                          " lvutil.create" % sr_uuid)
```
chandrikas: Could we simplify this to "No space to claim on a full SR %s"?

tests/test_trim_util.py
```diff
@@ -242,6 +256,7 @@ def test_do_trim_when_trim_succeeded_returns_true(self,
                                                       MockLock,
                                                       lvutil,
                                                       pread2):
+        lvutil._getVGstats.return_value = {'physical_size': 0, 'physical_utilisation': 0, 'freespace': EMPTY_VG_SPACE}
```
chandrikas: Line too long

chandrikas:
Looks good. Please add it to the patchqueue

PR #329, opened 2016-08-04
CA-209401: add static-vdi detach feature for HA

zli:
These are SM changes needed for this.

### DIFF
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -81,6 +81,7 @@ class NFSSR(FileSR.FileSR):
             self.sm_config = self.session.xenapi.SR.get_sm_config(self.sr_ref)
         else:
             self.sm_config = self.srcmd.params.get('sr_sm_config') or {}
+            self.other_config = self.srcmd.params.get('sr_other_config') or {}
         self.nosubdir = self.sm_config.get('nosubdir') == "true"
         if self.dconf.has_key('serverpath'):
             self.remotepath = os.path.join(self.dconf['serverpath'],
@@ -273,6 +274,7 @@ class NFSFileVDI(FileSR.FileVDI):
         resp['sr_uuid'] = sr_uuid
         resp['vdi_uuid'] = vdi_uuid
         resp['sr_sm_config'] = self.sr.sm_config
+        resp['sr_other_config'] = self.sr.other_config
         resp['command'] = 'vdi_attach_from_config'
         # Return the 'config' encoded within a normal XMLRPC response so that
         # we can use the regular response/error parsing code.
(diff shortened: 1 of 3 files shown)


zli:
The second part is in xapi, which currently exists as  https://github.com/zli/xen-api/tree/CA-209401.

The xapi changes can only get in after the SM changes are in place.

These two parts have been combine tested with most HA test sequences in nightly and regression suites.

drivers/SRCommand.py
```diff
@@ -299,6 +299,14 @@ def _run(self, sr, target):
             finally:
                 os.unsetenv('THIN_STATE_FILE_ATTACH')
 
+        elif self.cmd == 'vdi_detach_from_config':
+            extras= {}
+            if target.sr.driver_config.get("ATTACH_FROM_CONFIG_WITH_TAPDISK"):
+                target = blktap2.VDI(self.vdi_uuid, target, self.driver_info)
+                extras['deactivate'] = True
+                extras['caching_params'] = caching_params
+            target.detach(self.params['sr_uuid'], self.vdi_uuid, **extras)
```
chandrikas: Shouldn't it be a no-op if the sr driver config doesn't have "ATTACH_FROM_CONFIG_WITH_TAPDISK"?

zli: From my code reading, the flag ``ATTACH_FROM_CONFIG_WITH_TAPDISK`` doesn't mean whether ``attach/detach_from_config`` is supported, rather it means whether tapdisk should be involved in the attach/detach process.

This is corresponding to the logic in the existing ``attach_from_config`` definition right above --- where when such flag is present, the target is re-instantiated by ``blktap2.VDI(...)`` before the attach, otherwise the attach method will still be called, just on plain VDI object instead.

drivers/NFSSR.py
```diff
@@ -273,6 +274,7 @@ def generate_config(self, sr_uuid, vdi_uuid):
         resp['sr_uuid'] = sr_uuid
         resp['vdi_uuid'] = vdi_uuid
         resp['sr_sm_config'] = self.sr.sm_config
+        resp['sr_other_config'] = self.sr.other_config
```
chandrikas: Why is other_config a requisite for NFSSR attach? Also, this throws a run-time exception. There is no self.sr.other_config

zli: Because the NFS mount implementation in nfs.py seems to depend on it now. Note that it's not the nfs.py version here, it's the version after applying your internal patch queue.

chandrikas: We usually take care of those aspects when we convert a PR to a patch in the patchqueue. So how should this be tested? With the patchqueue applied or not?

zli: Fair enough.  The version here is better to test with patch queue applied, and it should look consistent at that point. Or you can use my working branch, which has it done automatically.

chandrikas:
Manual tests were good. I'll get these patches into the patchqueue to undergo a few more tests.

PR #331, opened 2016-08-12
CA-210336: Added exception handling to prevent stack trace dump

letsboogey:
Added a try-except block around line 353 which lists devices
at a given path in BaseISCSI.py attach method. Also made change
in SRCommand.py to prevent raising exceptions that are not of
type SR.SRException and dumping the stack trace on command line.


drivers/BaseISCSI.py
```diff
@@ -350,7 +350,12 @@ def attach(self, sr_uuid):
         if self.dconf.has_key('SCSIid'):
             if self.mpath == 'true':
                 self.mpathmodule.refresh(self.dconf['SCSIid'], 0)
-            devs = os.listdir("/dev/disk/by-scsid/%s" % self.dconf['SCSIid'])
+            try:
+                devs_path = "/dev/disk/by-scsid/%s" % self.dconf['SCSIid']
+                devs = os.listdir(devs_path)
+            except:
+                util.SMlog("[Errno 2] No such file or directory: '%s'" % devs_path)
```
MarkSymsCtx: What happens after the exception is handled here, the code blindly continues which is probably not what we want here.

letsboogey: I have raised a XenError exception and the most accurate error code I could find for this exception is the one I pasted below. Does that look right?
` <code>
                <name>ConfigDeviceInvalid</name>
                <description>The device is not a valid path</description>
                <value>91</value>
        </code>
`

MarkSymsCtx: That looks fine

PR #333, opened 2016-08-19
CA-218953: Use other-config parameter to identify tools repository

chandrikas:
We currently assume that any legacy_mode ISO SR houses guest tools
VDIs. This patch removes that assumption and relies on the
other-config:xenserver_tools_sr instead.


germanop:
Ack

PR #334, opened 2016-09-05
CA-198309: Cleanup locking files after inline vdi delete

stefanopanella:
This is adding the necessary locking files cleanup which was missed
during commit 9d82cecfe7a4a65bfdcd451235587c449c675923 ([CA-183960]
[CA-196882]: GC in batch mode and immediate deletes in SRs).

This has been tested on EXT and LVM local SR, where LVM local SR was
the environment used when the resource leak was found.

A further refactoring of the file locking lifecycle could be considered
but could also add significant risk which maybe we do not want to take
at this time.

Signed-off-by: Stefano Panella <stefano.panella@citrix.com>


MarkSymsCtx:
Looks good

PR #335, opened 2016-09-19
CA-219124: Wait for NFS service to be up before attempting mount

chandrikas:
Added wait (maximum of 3 minutes) for availability of NFS service
before attempting mount of NFS SR. Without this an NFS mount fails
immediately and an absence of retries leaves the NFS SR in a failed state. 


MarkSymsCtx:
Change looks ok. It might be an idea to add a unit test to assert that we call rpcinfo, basically the same as the modified tests but checking the other executable is called. Will make sure that we don't introduce any regressions later.

chandrikas:
Sure, will add those.

PR #336, opened 2016-10-07
PAR-154: Multipath path count misbehaves in port failure/restore in scalability test

swathicitrix:
Signed-off-by: Swathi Kovvuri swathi.kovvuri@citrix.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,6 +202,7 @@ install: precheck
 	ln -sf $(SM_DEST)lcache.py $(SM_STAGING)$(BIN_DEST)tapdisk-cache-stats
 	ln -sf /dev/null $(SM_STAGING)$(UDEV_RULES_DIR)/69-dm-lvm-metad.rules
 	install -m 755 scripts/xs-mpath-scsidev.sh $(SM_STAGING)$(UDEV_SCRIPTS_DIR)
+	install -m 755 scripts/udevmpath.sh $(SM_STAGING)$(UDEV_SCRIPTS_DIR)
 	install -m 755 scripts/xe-get-arrayid-lunnum $(SM_STAGING)$(BIN_DEST)
 	install -m 755 scripts/xe-getarrayidentifier $(SM_STAGING)$(BIN_DEST)
 	install -m 755 scripts/xe-getlunidentifier $(SM_STAGING)$(BIN_DEST)
--- a/drivers/mpathcount.py
+++ b/drivers/mpathcount.py
@@ -33,12 +33,13 @@ LOCK_NS2 = "mpathcount2"
 MP_INUSEDIR = "/dev/disk/mpInuse"
 mpp_path_update = False
 match_bySCSIid = False
+MPATHCOUNT_WAIT_INTERVAL = 2 
+
 
 util.daemon()
-if len(sys.argv) > 1:
+if len(sys.argv) == 3:
     match_bySCSIid = True
     SCSIid = sys.argv[1]
-if len(sys.argv) > 2:
     mpp_path_update = True
     mpp_entry = sys.argv[2]
 
@@ -63,6 +64,8 @@ elif not mpathcountlock.acquireNoblock():
     mpathcountqueue.release()
 
 util.SMlog("MPATH: I get the lock")
+time.sleep(MPATHCOUNT_WAIT_INTERVAL)
+util.SMlog("MPATH: Time wait is done")
 
 cached_DM_maj = None
 def get_dm_major():
--- a/udev/40-multipath.rules
+++ b/udev/40-multipath.rules
@@ -7,5 +7,5 @@ ACTION=="change", PROGRAM!="/sbin/dmsetup info -c --noheadings -j %M -m %m", GOT
 PROGRAM!="/sbin/dmsetup info -c -o uuid,name --separator ' ' --noheadings -j %M -m %m", GOTO="end_mpath"
 RESULT!="mpath-*", GOTO="end_mpath"
 # ENV is necessary otherwise the child process of mpathcount cannot find multipathd executable
-ACTION=="change", ENV{SCSIID}="%c{2}", ENV{PATH}="/sbin:/bin:/usr/sbin:/usr/bin", RUN+="/opt/xensource/sm/mpathcount.py $env{SCSIID}"
+ACTION=="change", ENV{SCSIID}="%c{2}", ENV{PATH}="/sbin:/bin:/usr/sbin:/usr/bin", RUN+="/etc/udev/scripts/udevmpath.sh"
 LABEL="end_mpath"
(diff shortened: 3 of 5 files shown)


udev/40-multipath.rules
```diff
@@ -7,5 +7,5 @@ ACTION=="change", PROGRAM!="/sbin/dmsetup info -c --noheadings -j %M -m %m", GOT
 PROGRAM!="/sbin/dmsetup info -c -o uuid,name --separator ' ' --noheadings -j %M -m %m", GOTO="end_mpath"
 RESULT!="mpath-*", GOTO="end_mpath"
 # ENV is necessary otherwise the child process of mpathcount cannot find multipathd executable
-ACTION=="change", ENV{PATH}="/sbin:/bin:/usr/sbin:/usr/bin", RUN+="/opt/xensource/sm/mpathcount.py %c{2}"
+ACTION=="change", ENV{SCSIID}="%c{2}", ENV{PATH}="/sbin:/bin:/usr/sbin:/usr/bin", RUN+="/opt/xensource/sm/udev_mpath.py $env{SCSIID}"
```
germanop: This one is conflicting with Zheng's patch that is already accepted, even though I did not push it yet here.

drivers/mpathcountdaemon.py
```diff
@@ -0,0 +1,63 @@
...
+
+import util
+import time, os, sys, re
+import xs_errors
+import glob
+import socket
+import sys
+import os
+
+server_address = "/uds_socket"
```
germanop: Root directory is not the right place for a socket

udev/40-multipath.rules
swathicitrix: As of now in my workspace,  on the master branch, Zheng changes are not available. I actually took the change done
by ZhengLi from Patch Queue and modified the file name.

Makefile
```diff
@@ -42,6 +42,8 @@ SM_LIBS += mpath_cli
 SM_LIBS += mpathutil
 SM_LIBS += LUNperVDI
 SM_LIBS += mpathcount
+SM_LIBS += mpathcountdaemon
```
swathicitrix: Done.

drivers/mpath_dmp.py
```diff
@@ -42,6 +42,7 @@
 DEVBYMPPPATH = "/dev/disk/by-mpp"
 SYSFS_PATH='/sys/class/scsi_host'
 UMPD_PATH='/var/run/updatempppathd.py.pid'
+MPATH_COUNT='/var/run/mpathcountdaemon.py.pid'
```
swathicitrix: Modified.

drivers/mpath_dmp.py
```diff
@@ -245,6 +246,10 @@ def activate():
     if not _is_mpp_daemon_running():
         cmd = ["service", "updatempppathd", "start"]
         util.pread2(cmd)
+        if not os.path.exists(MPATH_COUNT):
```
swathicitrix: Modified.

swathicitrix: I modified the code as per your comments.

drivers/mpathcount.py
```diff
@@ -36,10 +36,10 @@
 
 util.daemon()
 if len(sys.argv) == 3:
-    match_bySCSIid = True
```
swathicitrix: Done.

drivers/RawHBASR.py
```diff
@@ -373,7 +373,7 @@ def attach(self, sr_uuid, vdi_uuid):
             # The SCSIid is already stored inside SR sm_config.
             # We need only to trigger mpathcount
             try:
-                cmd = ['/opt/xensource/sm/mpathcount.py', scsi_id]
+                cmd = ['python','/opt/xensource/sm/udev_mpath.py',scsi_id]
```
swathicitrix: Done.

drivers/SR.py
```diff
@@ -486,7 +486,7 @@ def _setMultipathableFlag(self, SCSIid=''):
             self.session.xenapi.SR.set_sm_config(self.sr_ref, sm_config)
 
             if self.mpath == "true" and len(SCSIid):
-                cmd = ['/opt/xensource/sm/mpathcount.py',SCSIid]
+                cmd = ['python','/opt/xensource/sm/udev_mpath.py',SCSIid]
```
swathicitrix: done.

drivers/mpathcountdaemon.py
swathicitrix: Modified and now place it in/var/run/uds_socket

udev/40-multipath.rules
germanop: I pushed all the outstanding patches, so Zheg's should be there.

germanop:
This patch need rebase. It is against an old version of sm

germanop:
I will leave specific errors aside for this time because we spent much time on those and I lost the overall picture (but just give a look at udev_mpath, there are a few in the first lines).

This implementation does not scale: the 2 most notable problems are:
* it does use an oriented socket (stream instead of dgram) that is not necessary
* the daemon has only one thread for listening and serving the requests

If you test it quickly you will see that it cannot sustain a heavy load.
In my test:
* I used a custom program (instead of mpathcount.py) that just print something and sleep for 20 seconds
* started mpathcountd manually
* run from a shell udev_mpath (make sure to pass something because your implementation crashed otherwise) repeatedly.

Well, you will see that after the 4th the next invocation of udev_mpath will get stuck (so, in real life, udev will get stuck and we don't want that).
This is because the daemon is listening (not needed if dgram but this is not the main problem), accepting requests and invoking the worker program in the same thread.
The first few invocations of udev_mpath return immediately just because the system is queuing your requests (even though you passed "1" to listen()) but the daemon is not processing them.
In fact, it is still busy waiting for the first mpathcount to return.
When the fifth arrives, the daemon is not ready, the OS is not queuing any more for you and your call get stuck. 

The daemon implementation had to be designed more carefully.

swathicitrix:
Germano,
Thanks for your valuable inputs here.
Initially, I was thinking of the limit for requests that can be queued through the socket IPC.
But somehow I missed in analyzing it further.

Below is a draft of my next approach. Let me know your views on this.

Step1: As of now socket calls used are  blocking calls. So need to make it non-blocking.
Step2: Stop calling mpathcount.py from mpathcountd, instead store this in any suitable data structure.
Step3: Create another daemon thread which can access the info stored by mpathcountd in specified time intervals  and call mpathcount.py accordingly.

stefanopanella:
It all looks good to me now.
Reviewed-by: Stefano Panella

germanop:
I will fix this line in the commit message when pushing if not fixed earlier
```
Reverted the changes in mpathcount.py to operate with specific *iscsiid*
```
scsid instead of iscsiid

germanop:
Superseded by #344

PR #338, opened 2016-10-24
CA-221768: Use appropriate mount options based on ISO type

chandrikas:
Refactored ISO SR attaching logic to correctly detect ISO
type and use appropriate mount options

Signed-off-by: Chandrika Srinivasan chandrika.srinivasan@citrix.com

### DIFF
--- a/drivers/ISOSR.py
+++ b/drivers/ISOSR.py
@@ -248,7 +248,8 @@ class ISOSR(SR.SR):
         mountcmd=[]
         location = util.to_plain_string(self.dconf['location'])
         self.credentials = os.path.join("/tmp", util.gen_uuid())
-        protocol = 'nfs'
+        # TODO: Have XC standardise iso type string
+        protocol = 'nfs_iso'
         options = ''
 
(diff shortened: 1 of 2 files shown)


drivers/ISOSR.py
```diff
@@ -247,31 +247,36 @@ def attach(self, sr_uuid):
...
-
         self.credentials = os.path.join("/tmp", util.gen_uuid())
+        protocol = 'nfs'
+
         if self.dconf.has_key('type'):
-            # Creation via XC
-            if self.dconf['type']=='cifs':
-                # CIFS specific stuff
+            protocol = self.dconf['type']
+        elif ":/" not in location:
```
MarkSymsCtx: Does the CIFS location not contain :// so will match this?

chandrikas: No SMB SRs always have share name starting with "\\" or "//"

MarkSymsCtx: Ah, right, yes, that's fine then

PR #340, opened 2016-10-31
CA-226070 Switched on logging for snapwatchd daemon

letsboogey:
Some XenRT tescases depend on snapwatchd daemon logs
in SMLog

### DIFF
--- a/snapwatchd/snapwatchd.service
+++ b/snapwatchd/snapwatchd.service
@@ -4,7 +4,7 @@ After=syslog.target network.target
 
 [Service]
 Type=simple
-ExecStart=/opt/xensource/sm/snapwatchd/snapwatchd -f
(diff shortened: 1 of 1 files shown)


chandrikas:
@letsboogey , we don't want to turn logging on by default. snapwatchd takes an additional parameter that conditionally turns logging on. We need to make sure snapwatchd is started with that option.

snapwatchd/snapdebug.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+i#!/usr/bin/python
```
MarkSymsCtx: unintended change here as well.

letsboogey: oops, will fix that

chandrikas:
Looks good. I'll get it into the patchqueue.

PR #344, opened 2016-12-13
Fix different multipath related issues (CP-19247 and CA-234893)

stefanopanella:
This pr should supercede ps 336 since it is fixing some problems found with pr 336

Revert "CA-208229: enable mpathcount filtering by SCSIid"
CP-19247: Fix mpathcount problems affecting scalability test
CA-234893: sm-multipath.service goes inactive after SM RPM upgrade

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,6 +157,8 @@ install: precheck
 	  $(SM_STAGING)/$(SYSTEMD_SERVICE_DIR)
 	install -m 644 etc/make-dummy-sr.service \
 	  $(SM_STAGING)/$(SYSTEMD_SERVICE_DIR)
+	install -m 644 multipath/mpcount.service \
+	  $(SM_STAGING)/$(SYSTEMD_SERVICE_DIR)
 	install -m 644 snapwatchd/snapwatchd.service \
 	  $(SM_STAGING)/$(SYSTEMD_SERVICE_DIR)
 	for i in $(UDEV_RULES); do \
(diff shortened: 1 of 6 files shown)


etc/mpcount.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
```
germanop: If my previous comment about etc gets accepted, rename this one

etc/mpcount.service
```diff
@@ -0,0 +1,6 @@
...
+Description=Multipath Count Service
+
+[Service]
+Unit=forking
```
germanop: Not an expert of systemd at all but should not this be Type=forking?

scripts/udevmpath.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo /opt/xensource/sm/mpathcount.py | at now 2> /dev/null
+exit 0
```
germanop: Given I already asked other changes.. missing newline here

Makefile
```diff
@@ -157,6 +157,8 @@ install: precheck
 	  $(SM_STAGING)/$(SYSTEMD_SERVICE_DIR)
 	install -m 644 etc/make-dummy-sr.service \
 	  $(SM_STAGING)/$(SYSTEMD_SERVICE_DIR)
+	install -m 644 etc/mpcount.service \
```
stefanopanella: OK, can you please suggest how should I change the Makefile to do that?

etc/mpcount.service
stefanopanella: as before, where should I put the file? just under multipath dir instead of etc?

etc/mpcount.service
stefanopanella: Yes, good catch, I do not know how that ended up like this

scripts/udevmpath.sh
stefanopanella: ok I will add a new line

PR #347, opened 2017-03-01
Fix trivial type mismatch

cheese:
Signed-off-by: Robin Lee <cheeselee@fedoraproject.org>

### DIFF
--- a/drivers/OCFSSR.py
+++ b/drivers/OCFSSR.py
@@ -91,7 +91,7 @@ class OCFSSR(FileSR.FileSR):
         except util.CommandException, inst:
              raise xs_errors.XenError('OCFSMount', 
                                       opterr='Failed to mount FS. Errno is %d' \
-                                             % os.strerror(inst.code))
+                                             % inst.code)
 
     def attach(self, sr_uuid):
         if not self._checkmount():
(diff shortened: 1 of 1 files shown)


drivers/OCFSSR.py
```diff
@@ -120,7 +120,7 @@ def detach(self, sr_uuid):
         except util.CommandException, inst:
                 raise xs_errors.XenError('OCFSUnMount', \
                       opterr='Failed to umount FS. Errno is %d' % \
-                      os.strerror(inst.code))
+                      inst.code)
```
MarkSymsCtx: I'd lean to actually printing the stringified error message (or possibly both).

Also AFAIK, this code isn't actually used - it certainly isn't in the Citrix commercial product.

cheese: Since inst.code is the exit status from subprocess, it is not generally mapped to string message by os.strerror.

cheese:
Fixed the other two occurrences.

drivers/OCFSSR.py
MarkSymsCtx: If these aren't actually errnos then the message string is also incorrect. Replace Errno with Error is probably best.

PR #348, opened 2017-03-13
CA-246343: prevent incorrect usage of getSCSIid

germanop:
getSCSIid is called sometimes with only the device basename.
If this is the case, adding '/dev/' in front of the name
should rectify this.

Removed the fallback case (legacy of a transitional phase while
updagrading the user space stack) that was just creating
confusion.

Signed-off-by: Germano Percossi \<germano.percossi@citrix.com\>


germanop:
I will fix the typo in the commit message

PR #355, opened 2017-12-06
Delete LUN on detach of RawISCSI

cardoe:
iscsi: Delete LUN on detach of RawISCSI

We need to evict stale cached entry for LUN that's undergoing migration.

Signed-off-by: Rick Mesta <rick.mesta@rackspace.com>
Signed-off-by: Doug Goldstein <doug.goldstein@rackspace.com>
Reviewed-by: Doug Goldstein <doug.goldstein@rackspace.com>

### DIFF
--- a/drivers/iscsilib.py
+++ b/drivers/iscsilib.py
@@ -303,6 +303,14 @@ def logout(portal, target, all=False):
     except:
         raise xs_errors.XenError('ISCSILogout')
 
+def delete(target):
+    cmd = ["iscsiadm", "-m", "node", "-o", "delete", "-T", target]
+    failuremessage = "Failed to delete target"
+    try:
+        (stdout,stderr) = exn_on_failure(cmd,failuremessage)
+    except:
+        raise xs_errors.XenError('ISCSIDelete')
+
 def get_luns(targetIQN, portal):
     refresh_luns(targetIQN, portal)
     luns=[]
--- a/drivers/BaseISCSI.py
+++ b/drivers/BaseISCSI.py
@@ -408,7 +408,7 @@ class BaseISCSISR(SR.SR):
                 realdev = os.path.realpath("/dev/disk/by-scsid/%s/%s" % (self.dconf['SCSIid'], dev))
                 util.set_scheduler(realdev.split("/")[-1], "noop")
 
-    def detach(self, sr_uuid):
+    def detach(self, sr_uuid, delete=False):
         keys = []
         pbdref = None
         try:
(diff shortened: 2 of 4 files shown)


cardoe:
We've seen an issue where a volume gets detached via nova and migrated to a new storage backend and then re-attached to the same guest. Because of this code that this PR removes it caches the old info and results in the re-attached volume not working. This causes it to rescan every time there is an attach and not use any cached data.

cardoe:
There are concerns about the side effects of this with large installations with a large amount of volumes so if you have any advice on how to best proceed with fixing the issue while retaining the performance that would be most appreciated.

cardoe:
So maybe a better a approach here would be to some how clear the item out of the cache on a detach?

BobBall:
This affects the Base SR type which includes all types using iSCSI.  Is there a fix that could be made in the RawISCSI sr type?

BobBall:
This appears to be a special case for the RawISCSI SR when a volume is migrated from one host to another. With LVMoISCSI we require the lun to be plugged to all hosts all the time so it doesn't migrated around in this way necessitating running discovery every time we attach, running only if we don't already have a record for the IQN should be sufficient.

As such can we isolate this change to the RawISCSI SR type by having a new method on the BaseISCSI class to discover the map, with the existing behaviour, and override the behaviour in the RawISCSI class?

cardoe:
@BobBall Rick Mesta took this and updated it based on the feedback you provided and based on some internal feedback. He has validated it with XenServer 7.1.1 and shown me his test results that I have validated. Please consider this updated version for inclusion.

drivers/iscsilib.py
```diff
@@ -303,6 +303,14 @@ def logout(portal, target, all=False):
     except:
         raise xs_errors.XenError('ISCSILogout')
 
+def delete(target):
+    cmd = ["iscsiadm", "-m", "node", "-o", "delete", "-T", target]
+    failuremessage = "Failed to delete target"
+    try:
+        (stdout,stderr) = exn_on_failure(cmd,failuremessage)
+    except:
+        raise xs_errors.XenError('ISCSIDelete')
```
BobBall: As this is trying to clean out a cache and is only(?) an issue on migration, should the delete failing always be considered fatal to the detach operation? I guess the alternative would be to log the failure - but I'm just a little nervous that there might be other scenarios where perhaps the delete does fail and we're now making detach fail in that case?

MarkSymsCtx: I think with a small refactoring we can make this entirely the problem of RawISCSISR and be 100% confident that it won't impact the code used by other SR types.

BobBall: I'm not nervous about other SR types - I think the current code is perfectly safe for LVMoISCSI and friends.  My concern is around RawISCSI use cases that don't involve migration and whether a failed delete should always be fatal

BobBall:
Actually, I think that's a fair point - the chances of a delete failing if the logout had succeeded is very slim.

MarkSymsCtx:
Merged externally in https://github.com/xapi-project/sm/commit/bcbb92bd665861a739a7f901a819b66339fda98f

PR #384, opened 2018-06-11
CA-290491: do not leaf coalesce caching VDIs

zli:
There is some bad interactions between caching VDIs (currently attached)
and SMGC's leaf coalescing process (which didn't take that into consideration).

This wasn't an issue before the change of CP-18840, when the leaf coalescing
was completely disabled on file-based SR.

This change still allows leaf coalescing on file-based SR as CP-18840, but just
avoids online leaf coalescing any caching VDIs. There is very little benefit to
do this (for file-based SRs), and offline coalscing these nodes is still possible
as soon as these VDIs become detached.

Signed-off-by: Zheng Li <zheng.li3@citrix.com>

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -414,6 +414,7 @@ class VDI:
     DB_VDI_TYPE = "vdi_type"
     DB_VHD_BLOCKS = "vhd-blocks"
     DB_VDI_PAUSED = "paused"
+    DB_CACHING = "caching"
     DB_GC = "gc"
     DB_COALESCE = "coalesce"
     DB_LEAFCLSC = "leaf-coalesce" # config key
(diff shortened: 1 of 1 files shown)


minli1:
Hi, @zli May I just ask questions about the `caching` VDI.
1. A caching VDI is a `currently-attached` VDI? 
2. So from your code a `caching` VDI we can get `caching` field in xapi `VDI` object. Right?
Thanks.

drivers/cleanup.py
```diff
@@ -1501,6 +1503,9 @@ def findLeafCoalesceable(self):
             if vdi.getConfig(vdi.DB_ONBOOT) == vdi.ONBOOT_RESET:
                 Util.log("Skipping reset-on-boot %s" % vdi)
                 continue
+            if vdi.getConfig(vdi.DB_CACHING) == "true":
```
MarkSymsCtx: From discussion via email, this should probably check the `allowed-caching=true` property of the VDI and block leaf-coalesce of any VHD declared to having caching enabled. As caching will be disabled in either of these situations

- The VDI has no parent
- The VDI's parent also has a parent (i.e. the tree is taller than 2 nodes)

(See - https://github.com/xapi-project/sm/blob/master/drivers/blktap2.py#L1828)

Coalescing a VDI with `allow-caching = true` will violate these constraints.

zli: As per discussion by email, I'm not totally convinced which approach is better, but I've created another version for review & consideration (PR #385).

PR #398, opened 2018-08-24
CA-286144: udev script calling xe pusb-scan on boot timing out when X…

YarsinCitrix:
…API is not up yet

This fix changed the calling part from udev rules to a new scripts which is a wrapper of
the modified original scripts. The wrapper will have atd to start the original scripts,
so to detach from udev daemon, in case of long wait. The original scripts is modified
to first call xapi-wait-init-complete to ensure xapi is ready, in case of startup and
xapi is still initialising.

Signed-off-by: YarsinCitrix <yarsin.he@citrix.com>


YarsinCitrix:
@MarkSymsCtx , so do you think systemd way will be better? having the actual scripts to set as an service, and have systemd to start the scripts when at udev events?

YarsinCitrix:
also, is the no work at overload a batch rule rather than at rule?

MarkSymsCtx:
the way to do this is a OneShot, NoWait systemd service that is started non-blocking. Please discuss with Tim Smith as he's currently fixing a similarly broken piece of code for multipath and there will be some commonalities

PR #399, opened 2018-08-30
CA-295775 Fix handling of multipath events

TimSmithCtx:
Sending multipath UDEV events via a systemd service into atd does not
work very well at getting them processed; systemd drops attempts to
start any service which is already in the process of starting (which
loses information in the case of an event storm) while atd is a batch
processor designed to do things "at some vague time in the future which
could turn out to be never if the system load is higher than some
number".

A better solution is to use neither, and when we are scanning for
updates because of a UDEV event, use a simple wait/run lock pair to
rate-limit scans while ensuring that every event is followed by a scan
as soon as possible so its new state is detected.

We can now remove the 2 second sleep in drivers/mpathcount.py

In addition, the match_pathup function in drivers/mpathcount.py was not
matching the 'faulty' status in the multipath output and thus was not
updating the status when it changed. (Cosmetic: removed a tab from a log
line, log why the db update failed if it did fail).

In drivers/mpath_cli.py invoke multipathd by its full path because we
know it. Do not rely on PATH search for things like this; udev does
not supply the PATH we want and we should avoid requiring it.

### DIFF
--- a/scripts/onequeue.sh
+++ b/scripts/onequeue.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/bash
+exec </dev/null &>/dev/null
+
+## This is a script more general than just SM and may eventually want
+## to move to another package. Its purpose is to help with running
+## a process to act on some event while other events may be arriving.
+##
+## Invoke as
+##
+## /etc/udev/scripts/onequeue.sh <lockname> <command> [args...]
+##
+## The first argument is a name for a lock (not a pathname, just a name).
+## This will be used for a wait lock (LOCK.wait) and a run lock (LOCK.run).
+## The remaining arguments are the command to run and arguments for that
+## command. Every different command/args set which might need to run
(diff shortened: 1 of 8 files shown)


scripts/onequeue.sh
```diff
@@ -0,0 +1,71 @@
...
+        $PROGARGS
+        /usr/bin/logger -t onequeue -p daemon.notice "'$PROGARGS' done"
+    ) 10>"$LOCKDIR/$LOCK.run"
+}
+
+(
+	## Attempt to acquire an exclusive lock on fd 9 (LOCK.wait)
+    /usr/bin/flock -x -n 9
+    if [ $? -ne 0 ]; then
+		## If we didn't get it, someone is already waiting, so just exit.
```
MarkSymsCtx: Spaces & tabs?

PR #401, opened 2018-09-12
CA-286144: udev script calling xe pusb-scan on boot timing out when X…

YarsinCitrix:
…API is not up yet

This fix is trying to address the issue that when xapi is not yet ready,
the udev rules is triggered and possible hang.

The fix creates a new Python scripts as daemon service, waiting on a pipe.
The udev rules will invoke another script (the same one but with different
mode) to trigger the daemon script to try to trigger xapi usb-scan.

The implementation assumes the server side to be run as systemd daemon, and
it waits on a named pipe, which when the udev rules are triggered, the wrapper
script will write hence wake up the server and the server will further invoke
the script to do the actual work as child.

Signed-off-by: YarsinCitrix <yarsin.he@citrix.com>


scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+            self.logErr("script passed is not found")
+            raise error
+        
+    def _openRead(self):
+        """ server side open named pipe to read """
+        try:
+            # here to have write permission also is necessary to keep one writer
+            # to avoid pipe 'broken' in case of other closure.
+            self.fifo = os.open(self.pipePath, os.O_RDWR | os.O_NONBLOCK)
+            self.epollObj.register(self.fifo, select.EPOLLIN)
```
taoyongd: Doesn't this need to unregister?

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+            raise err
+            
+    def work(self):
+        """ simply write the request and end ourselves """
+        self.logInfo("client started")
+        
+        try:
+            self._openWrite()
+            self._writeRequest()
+            self._closeFIFO()
```
MarkSymsCtx: This could be moved to a finally block to avoid duplication with the error case.

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+            self.logErr("Write error at:" + str(err))
+            raise err
+
+    def _openWrite(self):
+        """ open pipe for write """
+        try:
+            self.fifo = os.open(self.pipePath, os.O_WRONLY)
+        except OSError as err:
+            self.logErr("open file write error:" + str(err))
+            raise err
```
MarkSymsCtx: just raise

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+    def _writeRequest(self):
+        """ write the pipe to wake up the daemon to do the work """
+        if self.fifo == None:
+            self._openWrite()
+            
+        try:
+            os.write(self.fifo, self.name)
+        except OSError as err:
+            self.logErr("Write error at:" + str(err))
+            raise err
```
MarkSymsCtx: just raise

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+    def __init__(self, name, script):
+        super(ServerInstance, self).__init__(name)
+        self.script = script
+        self.epollObj = select.epoll()
+        
+        # check if the script to be invoked exists
+        subargs = script.split(' ')
+        if not os.path.isfile(subargs[0]):
+            self.logErr("script passed is not found")
+            raise error
```
MarkSymsCtx: what is error? You need to specify an exception to raise here

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _checkPath(self):
+        """ check if the paths and files exist, otherwise create them """
+        
+        if not os.path.exists(self.PIPE_BASE_PATH):
+            try:
+                os.mkdir(self.PIPE_BASE_PATH, 0755)
+            except OSError as err:
+                self.logErr("create base path failed!")
+                raise err
```
MarkSymsCtx: just raise

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+            As we hold one writer, there should be no EPOLLHUP when others closed the pipe.
+        """
+        events = self.epollObj.poll(-1)
+        for fileno, event in events:
+            if fileno == self.fifo and event & select.EPOLLIN:
+                return True
+            elif fileno == self.fifo and event & (select.EPOLLERR | select.EPOLLHUP):
+                # probably some log and raise failure
+                self.logErr("exit due to unexpected events:" + str(event))
+                raise OSError
```
MarkSymsCtx: OSError will need some arguments here

scripts/service.py
```diff
@@ -0,0 +1,247 @@
...
+    service = args.service
+    mode = args.mode
+    executable = args.executable
+    
+    if mode == 'client':
+        inst = ClientInstance(service)
+    else: # server
+        inst = ServerInstance(service, executable)
+    
+    inst.work()
```
MarkSymsCtx: No newline at file end

YarsinCitrix:
@MarkSymsCtx , and @TimSmithCtx , I have updated the code according to previous review comments and also added unit test. Please help to check this out!

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+                create_fifo = True
+        else:
+                create_fifo = True
+
+        if create_fifo:
+            try:
+                os.mkfifo(self.pipePath)
+            except OSError as err:
+                if err.errno != errno.EEXIST:
+                    self.loggErr("Error encoutnered while create pipe")
```
MarkSymsCtx: typo

YarsinCitrix:
@TimSmithCtx , can you please help to spare some time to help to check this out, as Mark added you in?

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+import xcp.logger as logger
+import errno
+import shutil
+import stat
+import sys
+
+BUFFER_SIZE = 2048
+
+class ServiceSkeleton(object):
+    """ base class for both client and server, maily as utilities """
```
minli1: typo `mainly` :)

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+
+    PIPE_BASE_PATH = '/var/lib/misc/'
```
TimSmithCtx: Put the pipes in /var/run/udevworker or something similar under /var/run, would be my advice. Since /var/run is a tmpfs, you are guaranteed not to have old rubbish lying around after a reboot, which makes everything a bit more predictable.

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+    def __init__(self, name):
+        super(ClientInstance, self).__init__(name)
+        
+    def _writeRequest(self):
+        """ write the pipe to wake up the daemon to do the work """
+        if self.fifo == None:
+            self._openWrite()
+            
+        try:
+            os.write(self.fifo, self.name)
```
TimSmithCtx: Since we actually care what is written through the pipe, only that a write happens, why write more than one byte? Takes longer to fill the pipe by accident that way.

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+        self.logInfo("will invoke scripts:" + self.script)
+        try:
+            subprocess.Popen(self.script)
+        except OSError as err:
+            self.logErr("execute script error:" + str(err))
+            
+    def _readRequest(self):
+        """ read all that is written in the pipe, so in case of input storm,
+            just do work once.
+        """
```
TimSmithCtx: This is the right idea, but there is a condition in which this goes wrong.

The reason we are running this as a service is that the script it invokes for us may take too long (either because it has to do a lot of work or because it just takes too long to communicate the results to xapi - which might be xapi's fault but we've still got to deal with it). When you are off in _executeSub() you are not reading the pipe, which means in the event of a storm clients can fill the pipe and block. This is fine from the server's perspective (it will correctly empty the pipe and invoke _executeSub() again) but lots of blocking clients may cause problems elsewhere in udev, delaying other scripts, running out of workers etc.

I fear you will need threads to solve that one :-(

YarsinCitrix: @TimSmithCtx , not sure if I understand correctly. But although what's actually executed in the _executeSub() may takes long, the main thread will not wait till it finishes. So the gap between the main thread leaves the pipe not read and the next read is only forking the sub process, should it should be very quick, and will not causing too many clients awaits there. It is probably not as you think the main will wait till the _executeSub() finishes and then back to read again.

scripts/service.py
YarsinCitrix: @TimSmithCtx , This actually leaves some space for further more trick check to verify the written is valid. Since the service name will be very short, so the time taken will not be different. The OS will do read/write per what's 'cached' in the buffer of the pipe, so written one bytes and several actually take the same time.

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+                    self.logErr("Unexpected file node at:" + self.pipePath)
+                    raise OSError(errno.EEXIST, "Unexpected file node exists", self.pipePath)
+
+                create_fifo = True
+        else:
+                create_fifo = True
+
+        if create_fifo:
+            try:
+                os.mkfifo(self.pipePath)
```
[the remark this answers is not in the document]
TimSmithCtx: Actually no. The worst case is udev killing those tasks after waiting some time and other event handlers which udev would have run not being run until that time has expired, or missed entirely as no workers were available. I have observed udev having this behaviour. systemd-udevd has a maximum number of children it will run, whose default is going to end up at around 8 per GB of RAM in Dom0 (src/udev/udevd.c in systemd). We can't rely on not hitting it if we block our workers too long - we are not the only thing udev is handling. 

From pipe(7): "If a process attempts to write to a full pipe (see below), then write(2) blocks until sufficient data has been read from the pipe to allow the write to complete."

So we have two cases where the client can block:

1) On open() when the FIFO exists but the server is not running
2) On write() when the FIFO is full, which can happen if a large number of events occurs while the server is off running the worker script.

1) can be avoided fairly easily, either by having that open() return ENOENT unless the server will be listening very soon (I mean milliseconds), or by doing a non-blocking open()
2) can be avoided by setting the opened fd non-blocking and ignoring EAGAIN and EPIPE from the write.

scripts/service.py
TimSmithCtx: I guess I read that wrong then. What stops more than one _executeSub() from running at once?

scripts/service.py
TimSmithCtx: You're not taking a lock on the FIFO when you open it for write. If udev fires off two of your workers in close succession called "myeventname", and you read "myevmyeventnameentname" in the server, how will you validate it? :-)

scripts/service.py
```diff
@@ -0,0 +1,286 @@
...
+        
+        try:
+            self._openRead()
+            while True:
+                if self._waitForInput():
+                    if self._readRequest():
+                        self._executeSub()
+        except Exception as err:
+            self.logErr("Exit at error:" + str(err))
+            self._closeFIFO()
```
minli1: do we need to `raise` here?

scripts/service.py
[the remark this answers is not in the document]
YarsinCitrix: @TimSmithCtx , OK, in that case, we can make the client part none blocking when at open.
For case 2, it is indeed that the server will not block waiting the child process to finish, so I guess we just do not need to take care of it.

scripts/service.py
YarsinCitrix: OK, I originally considered, but did not do just to avoid complicated init file/folder sequence.

scripts/service.py
YarsinCitrix: It is not kind of byte level partition. For PIPE, each write is done per one pipe buffer, which should be 4K or 64K bytes (also in that case, one byte or several bytes does not have performance difference), so if we just use some small string as service name, it wouldn't get split/mixed. But any way, it is just a potential, and we can leave it there.

scripts/service.py
YarsinCitrix: Yes, the subprocess will only fork the child, and only return a stub to the child process, and we did not do any extra wait/check, so just after fork, the server returns, none blockingly.

In the situation of this usb scan case, my understanding is to not miss any lasted events. So if we just skip the invokation if there is one already, the latest change might be missing (supposing the current one is in some final clean up stage). 

But if this service is made general, there might be such case that only one active job is expected. So I added an option from the server side to define the behavior, as one child only, or multiple.

scripts/service.py
YarsinCitrix: Not necessarily, from the main logic, after the work ends, the process/thread will exit.

scripts/service.py
TimSmithCtx: write() is not atomic, so if you write more than 1 byte at a time from 2 processes to the same file/pipe without locking, at some point they *will* get mixed. It's not important now but would be if we tried to use that.

YarsinCitrix: @TimSmithCtx , yes, write() itself is not atomic, but this is write to a pipe. Please refer to the following for details, as what I described above:
From http://man7.org/linux/man-pages/man7/pipe.7.html:
==========================
POSIX.1 says that write(2)s of less than PIPE_BUF bytes must be
       atomic: the output data is written to the pipe as a contiguous
       sequence.  Writes of more than PIPE_BUF bytes may be nonatomic: the
       kernel may interleave the data with data written by other processes.
       POSIX.1 requires PIPE_BUF to be at least 512 bytes.  (On Linux,
       PIPE_BUF is 4096 bytes.) 
==========================
But anyway, we still did not do that, so let's just not wasting more time discussing this.

scripts/service.py
YarsinCitrix: For use case here,  there is no such issue. The child will only call xapi to scan usb. It is a simple cli call. The problem is the script might block, worst in several minutes. And I think we might also need to prepare for the worst case, a dead block in the child, as we do not know what's in the scripts. Still, I think this will be a case by case decision. If any child has risk in concurrent, I think it should also be addressed in the child itself, as we cannot control who will run the scripts and how. Also, making a server block is always the worse choice.

YarsinCitrix: @TimSmithCtx , For your mentioned implementation, you can start the server with '-u' option, so there is only one running child.

scripts/service.py
TimSmithCtx: Yes, fair enough.

scripts/service.py
TimSmithCtx: Yes that's the case we want.

My point is I think we'll never want any other cases, and if we do the solution will more properly be done elsewhere, but it doesn't seem worth holding this up.

YarsinCitrix: @TimSmithCtx , I think in the case here, multiple child/invoke is a better solution. Let me try to explain once again. Think of one udev (device add) event fires. So when the server is processing e1, child is triggered and is processing something, here to call the command to invoke xapi to do usbscan. Unknowing the timing and sequence, the xapi could quickly get triggered and processed, before the child returns completely. Then at the moment, after xapi complete the scanning and before the child returns completely, another udev event fires. In this case, if we just skip the invoking because the child is not yet returns, the event (for example new device added) will be missed from processing, which is not good.

TimSmithCtx: If the read side simply blocks until the child has returned completely, then any and all events which occurred after it was started will be sitting in the pipe, ready to be coalesced into a single "go again" instruction and trigger another run. The pipe might be full, but we don't care because the writers don't block.

TimSmithCtx: That's a bug in the child and should be fixed there. It would be sufficient to log that we started it. Worst case the faulty service stops working.

MarkSymsCtx:
Superseded by https://github.com/xapi-project/sm/pull/412 which is now merged.

PR #404, opened 2018-10-03
CP-28924: Declare thin provisioning capability for file based SRs

stefanopanella:
Signed-off-by: Thomas Mckelvey <thomas.mckelvey@citrix.com>


stefanopanella:
Thanks Mark for spotting that, we have just discussed with Robert that it would make sense to do same thing for local ext just because THIN_PROVISIONING is a hint to XenCenter  that there is no need to check for the available space in the SR since the VDI will just grow on demand. I will send an updated patch including ext SR.

stefanopanella:
Updated patch to include EXTSR.py

PR #412, opened 2018-11-05
CA-286144: add usb-scan systemd units

srowe:
udev does not tolerate the time it can take for XAPI to complete a
pUSB rescan request. Use a service listening on a socket to perform
the scan request so that the udev action just becomes writing to a
non-blocking pipe.


srowe:
This needs spec files changes to add and activate the new units.

PR #414, opened 2018-11-08
CP-27709: suppress error if blkdiscard fails on full provisioned lun

MarkSymsCtx:

### DIFF
--- a/drivers/lvutil.py
+++ b/drivers/lvutil.py
@@ -19,12 +19,14 @@
 #
 
 
+import glob
 import re
 import os
 import errno
 import time
 
 import util
+import scsiutil
 import xs_errors
 import xml.dom.minidom
 from lvhdutil import VG_LOCATION,VG_PREFIX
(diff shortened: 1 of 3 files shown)


drivers/scsiutil.py
```diff
@@ -182,9 +182,8 @@ def scsi_dev_ctrl(ids, cmd):
             opterr='An error occured during the scsi operation')
 
 def getdev(path):
-    realpath = os.path.realpath(path)
```
MarkSymsCtx: It turns out that calling realpath here meant that the match_dm check in the next line could never be true as any device of the form `/dev/mapper/xxxxx/` will realpath to something of the form `/dev/dm-xx`

MarkSymsCtx:
It now just logs to SMlog if a discard to a full provisioned LUN fails (in most cases we'd expect that it will but there may be some SANs out there that will accept a discard on a LUN marked full).

stefanopanella:
Lgtm

drivers/lvutil.py
```diff
@@ -725,3 +728,23 @@ def removeDevMapperEntry(path, strict=True):
         util.SMlog("removeDevMapperEntry: dmsetup remove failed for file %s " \
                    "with error %s, and lsof ret is %s." % (path, str(e), ret))
         return False
+
+def vgIsFullyProvisioned(vgname):
+    """
+    Check the PVs behind the specified VG and return
+    True if any of them is fully provisioned
+    """
+    text = cmd_lvm([CMD_VGDISPLAY, '-v', vgname])
```
chandrikas: Is there an easier way to find underlying devices? Can we use self.root?

chandrikas: hmm, I see that getting from lvutil to LVHDSR is going to be convoluted too. Let's go with this approach then.

PR #416, opened 2018-11-21
Revert "CP-27709: suppress error if blkdiscard fails on full provisio…

MarkSymsCtx:
…ned lun"

This reverts commit 47fabe7eadc16e674d3c4adba6daf6097abf9c03.

### DIFF
--- a/drivers/lvutil.py
+++ b/drivers/lvutil.py
@@ -19,14 +19,12 @@
 #
 
 
-import glob
 import re
 import os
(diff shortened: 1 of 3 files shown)


MarkSymsCtx:
Unfortunately the change to getdev causes LUN resize to fail consistently. Needs further investigation so back this change out for now.

stefanopanella:
ack

PR #420, opened 2019-01-02
Update MAX_VHD_SIZE to 2088960MiB (2040GiB).

GeraldElder-Vass:
Signed-off-by: Gerald Elder-Vass <gerald.elder-vass@citrix.com>

### DIFF
--- a/drivers/vhdutil.py
+++ b/drivers/vhdutil.py
@@ -28,7 +28,7 @@ import xs_errors
 import time
 
 MIN_VHD_SIZE = 2 * 1024 * 1024
-MAX_VHD_SIZE = 2093050 * 1024 * 1024
+MAX_VHD_SIZE = 2040 * 1024 * 1024 * 1024
 MAX_VHD_JOURNAL_SIZE = 6 * 1024 * 1024 # 2MB VHD block size, max 2TB VHD size
 MAX_CHAIN_SIZE = 30 # max VHD parent chain size
 VHD_UTIL = "/usr/bin/vhd-util"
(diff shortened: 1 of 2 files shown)


GeraldElder-Vass:
Relates to CA-300719

GeraldElder-Vass:
Can someone explain why changing the MAX_VHD_SIZE causes this failure? (To me it looks like something is missing instead of it failing to do something with the number I've provided?)

GeraldElder-Vass:
Can someone explain why changing the MAX_VHD_SIZE causes this failure? (To me it looks like something is missing instead of it failing to do something with the number I've provided?)

gaborigloi:
`test_validate_and_round_max_size` test has been failing:
https://github.com/xapi-project/sm/blob/master/tests/test_vhdutil.py#L18-L21

MarkSymsCtx:
The reason that it looks like something is missing is that the error handlers aren't mocked up properly, so when something fails it then fails to lookup the error details in the XML file.

drivers/vhdutil.py
```diff
@@ -28,7 +28,7 @@
 import time
 
 MIN_VHD_SIZE = 2 * 1024 * 1024
-MAX_VHD_SIZE = 2093050 * 1024 * 1024
+MAX_VHD_SIZE = 2088960 * 1024 * 1024
```
MarkSymsCtx: I wonder whether this would be clearer as

`2040 * 1024 * 1024 * 1024`

or even 

`GB_BYTES = 1024 * 1024 * 1024`
`MAX_VHD_SIZE = 2040 * GB_BYTES`

GeraldElder-Vass: I think  2040*1024*1024*1024 is more in line with the rest of this file, also GB_BYTES (gigabytes_bytes) might be confusing.

PR #436, opened 2019-03-15
CA-312608: Set scheduler to noop for tdX devices

rosslagerwall:
With newer kernels, changing the scheduler in-kernel shortly after
device creation can deadlock. Instead, have SM set the scheduler. This
kind of policy decision is better kept in userspace anyway.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/drivers/util.py
+++ b/drivers/util.py
@@ -977,6 +977,22 @@ def dom0_disks():
     SMlog("Dom0 disks: %s" % disks)
     return disks
 
+def set_scheduler_sysfs_node(node, str):
+    """Set the scheduler for a sysfs node (e.g. '/sys/block/sda')"""
+
+    path = os.path.join(node, "queue", "scheduler")
+    if not os.path.exists(path):
+        SMlog("no path %s" % path)
+        return
+    try:
+        f = open(path, 'w')
+        f.write("%s\n" % str)
+        f.close()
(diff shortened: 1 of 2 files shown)


drivers/util.py
```diff
@@ -977,6 +977,22 @@ def dom0_disks():
     SMlog("Dom0 disks: %s" % disks)
     return disks
 
+def set_scheduler_sysfs_node(node, str):
+    """Set the scheduler for a sysfs node (e.g. '/sys/block/sda')"""
```
MarkSymsCtx: This comment doesn't match the usage above `/sys/dev/block/x:x`, maybe have the two expected forms?

drivers/util.py
```diff
@@ -977,6 +977,22 @@ def dom0_disks():
...
+
+    path = node + "/queue/scheduler"
```
MarkSymsCtx: `os.path.join(node, "queue", "scheduler")` please

drivers/util.py
rosslagerwall: sysfs has many different symlinks pointing to the same sysfs node (`/sys/block/sda` itself is actually a symlink). I'm don't think that it is useful or necessary to list an extra form.

PR #443, opened 2019-05-14
Avoid getting exceptions logged from checking if an lv exists

MarkSymsCtx:
CBT code make many checks for the existence of the cbtlog volume but if CBT is not enabled this logs exceptions when the volume is not found. Use doexec directly and check if the return code is 0 or not.


chandrikas:
LGTM. Why does it show unrelated commits though?

MarkSymsCtx:
I hadn't spotted that it was doing, only looked at the file diff, that's weird. Will go look at the history/rebase.

MarkSymsCtx:
That looks more sane.

PR #452, opened 2019-07-23
Ca 323702: fix semantics of locking code

MarkSymsCtx:


drivers/lock.py
```diff
@@ -25,56 +25,39 @@
...
-                # If another lock within the namespace has already
-                # cleaned up the namespace by removing the directory,
-                # _open_lockfile raises an ENOENT, in this case we retry.
-                if e.errno == errno.ENOENT:
-                    if number_of_enoent_retries > 0:
-                        number_of_enoent_retries -= 1
-                        continue
-                raise
-            break
+    # These are required to pacify pylint as it doesn't understancd the __new__
```
BenSimsCitrix: s/understancd/understand/

drivers/lock.py
```diff
@@ -25,56 +25,39 @@
...
+    def acquire(self):
+        raise NotImplementedError("Lock methods implemented in LockImplementation")
 
-        fd = self.lockfile.fileno()
-        self.lock = flock.WriteLock(fd)
+    def acquireNoblock(self):
```
BenSimsCitrix: Curious why we need these, does the AttributeError error not contain enough info to debug this?

BenSimsCitrix: Unless we have an object deriving from this class that i have not got to yet :)

MarkSymsCtx: See the comment above. pylint claims it to be an error if the method doesn't exist. Because the code calls `lock = Lock('foo')` but actually gets a `LockImplementation` returned to it, which is not and cannot be a subclass of Lock (circular recursion in inheritance if you do that :() Decided this was less ugly than suppressing the pylint error here.

drivers/lock.py
```diff
@@ -163,20 +213,29 @@ def acquire(self):
...
 
     def acquireNoblock(self):
         """Acquire lock if possible, or return false if lock already held"""
-        exists = os.path.exists(self.lockpath)
-        ret = self.lock.trylock()
-        if VERBOSE:
-            util.SMlog("lock: tried lock %s, acquired: %s (exists: %s)" % \
-                    (self.lockpath, ret, exists))
+        if not self.held():
+            exists = os.path.exists(self.lockpath)
```
BenSimsCitrix: Move this line to bellow the 

if VERBOSE

MarkSymsCtx: errr, no, that's not what we want.

tests/test_lock.py
```diff
@@ -49,12 +57,95 @@ def test_lock_file_create_fails_retried(self, context):
...
+
+    @testlib.with_context
+    def test_lock_acquire_twice_noblock_release(self, context):
+        self.setup_fcntl_return(context)
+
+        lck1 = lock.Lock("somename")
+
+        lck1.acquire()
+
+        self.assertTrue(lck1.held())
```
BenSimsCitrix: Perhaps some value in copy pasting these tests twice more so that we get all four permutations.
aquire,aquire DONE
aquire, noblock DONE
noblock, aquire MISSING
noblock, noblock MISSING

drivers/lock.py
BenSimsCitrix: But the variable "exists" is only used in the log comments, unless i've missed something.

MarkSymsCtx: but it needs to be done before taking the lock as it's guaranteed to exist after taking the lock.

tests/test_lock.py
MarkSymsCtx: Done

drivers/lock.py
BenSimsCitrix: Dohh, understood

PR #456, opened 2019-09-23
CA-327382: reap child processes

MarkSymsCtx:
This caused XSI-474 and other reported customer issues.

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -147,22 +147,35 @@ class Util:
         pid = os.fork()
         if pid:
             startTime = time.time()
-            while True:
-                if resultFlag.test("success"):
-                    Util.log("  Child process completed successfully")
-                    resultFlag.clear("success")
-                    return
-                if resultFlag.test("failure"):
-                    resultFlag.clear("failure")
-                    raise util.SMException("Child process exited with error")
-                if abortTest() or abortSignaled:
-                    os.killpg(pid, signal.SIGKILL)
-                    raise AbortException("Aborting due to signal")
-                if timeOut and time.time() - startTime > timeOut:
-                    os.killpg(pid, signal.SIGKILL)
-                    resultFlag.clearAll()
-                    raise util.SMException("Timed out")
-                time.sleep(pollInterval)
+            try:
+                while True:
+                    if resultFlag.test("success"):
+                        Util.log("  Child process completed successfully")
+                        resultFlag.clear("success")
+                        return
+                    if resultFlag.test("failure"):
+                        resultFlag.clear("failure")
+                        raise util.SMException("Child process exited with error")
+                    if abortTest() or abortSignaled:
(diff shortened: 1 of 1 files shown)


drivers/cleanup.py
```diff
@@ -147,22 +147,35 @@ def runAbortable(func, ret, ns, abortTest, pollInterval, timeOut):
...
+                        raise AbortException("Aborting due to signal")
+                    if timeOut and time.time() - startTime > timeOut:
+                        os.killpg(pid, signal.SIGKILL)
+                        resultFlag.clearAll()
+                        raise util.SMException("Timed out")
+                    time.sleep(pollInterval)
+            finally:
+                wait_pid = 0
+                rc = -1
+                count = 0
```
BenSimsCitrix: Could we simplify the code by having a resultFlag.clearAll() in the finally?

MarkSymsCtx: Actually, the whole resultflag piece can be removed, but that's a bigger change for another day. We can just use the exit code from the process by moving the `waitpid(pid, WNOHANG)` inside the loop, basically poll the subprocess until it completes and the read the exit status.

drivers/cleanup.py
```diff
@@ -147,22 +147,35 @@ def runAbortable(func, ret, ns, abortTest, pollInterval, timeOut):
...
+                        resultFlag.clear("success")
+                        return
+                    if resultFlag.test("failure"):
+                        resultFlag.clear("failure")
+                        raise util.SMException("Child process exited with error")
+                    if abortTest() or abortSignaled:
+                        os.killpg(pid, signal.SIGKILL)
+                        raise AbortException("Aborting due to signal")
+                    if timeOut and time.time() - startTime > timeOut:
+                        os.killpg(pid, signal.SIGKILL)
```
BenSimsCitrix: perhaps we could tidyup by having a resultFlag.clearall() in the finally.

drivers/cleanup.py
```diff
@@ -147,22 +147,35 @@ def runAbortable(func, ret, ns, abortTest, pollInterval, timeOut):
...
+                    time.sleep(pollInterval)
+            finally:
+                wait_pid = 0
+                rc = -1
+                count = 0
+                while wait_pid == 0 and count < 10:
+                    wait_pid, rc = os.waitpid(pid, os.WNOHANG)
+                    if wait_pid == 0:
+                        time.sleep(2)
+                        count += 1
```
BenSimsCitrix: docs suggest that an OSError can be raised, do we want to catch it or bomb out.

drivers/cleanup.py
TimSmithCtx: > perhaps we could tidyup by having a resultFlag.clearall() in the finally.

Do we still care at that point? If we entered the finally: as the result of an exception, the exception will be re-thrown when the finally completes.

PR #463, opened 2019-11-01
CA-329845 - Remove usage of credentials file for CIF

BenSimsCitrix:
Signed-off-by: ben sims <ben.sims@citrix.com>

### DIFF
--- a/drivers/ISOSR.py
+++ b/drivers/ISOSR.py
@@ -21,6 +21,7 @@ import SR, VDI, SRCommand, util
 import nfs
 import os, re
 import xs_errors
+import cifutils
 
 CAPABILITIES = ["VDI_CREATE", "VDI_DELETE", "VDI_ATTACH", "VDI_DETACH", 
                 "SR_SCAN", "SR_ATTACH", "SR_DETACH"]
@@ -230,7 +231,6 @@ class ISOSR(SR.SR):
 
         # Some info we need:
         self.sr_vditype = 'phy'
-        self.credentials = None
 
     def delete(self, sr_uuid):
         pass
@@ -254,7 +254,6 @@ class ISOSR(SR.SR):
 
         mountcmd=[]
         location = util.to_plain_string(self.dconf['location'])
-        self.credentials = os.path.join("/tmp", util.gen_uuid())
         # TODO: Have XC standardise iso type string
         protocol = 'nfs_iso'
         options = ''
@@ -305,7 +304,6 @@ class ISOSR(SR.SR):
             # Check the validity of 'smbversion'.
             # Raise an exception for any invalid version.
             if self.smbversion not in [SMB_VERSION_1, SMB_VERSION_3]:
-                self._cleanupcredentials()
                 raise xs_errors.XenError('ISOInvalidSMBversion')
 
         # Attempt mounting
@@ -344,21 +342,18 @@ class ISOSR(SR.SR):
                                 mountcmd.extend(options)
                             self.mountOverSMB(mountcmd)
                         else:
-                            self._cleanupcredentials()
                             raise xs_errors.XenError(
                                 'ISOMountFailure', opterr=inst.reason)
                 else:
                     util.SMlog('ISOSR mount over smb 1.0')
                     self.mountOverSMB(mountcmd)
         except util.CommandException, inst:
-            self._cleanupcredentials()
             if not self.is_smbversion_specified:
                 raise xs_errors.XenError(
                     'ISOMountFailure', opterr=smb3_fail_reason)
             else:
                 raise xs_errors.XenError(
                     'ISOMountFailure', opterr=inst.reason)
-        self._cleanupcredentials()
 
         # Check the iso_path is accessible
         if not self._checkmount():
@@ -384,7 +379,10 @@ class ISOSR(SR.SR):
 
     def mountOverSMB(self, mountcmd):
         """This function raises util.CommandException"""
-        util.pread(mountcmd, True)
+        new_env, domain = cifutils.getCIFCredentials(self.dconf, self.session,
+                                                     prefix="cifs")
+
+        util.pread(mountcmd, True, new_env=new_env)
         try:
             if not self.is_smbversion_specified:
                 # Store the successful smb version in PBD config
@@ -421,10 +419,19 @@ class ISOSR(SR.SR):
         """Append options to mount.cifs"""
         options = []
         try:
-            options.append(self.getCIFSPasswordOptions())
             options.append(self.getCacheOptions())
-            options.append('guest')
+
+            if not cifutils.containsCredentials(self.dconf, prefix="cifs"):
+                options.append('guest')
+
             options.append(self.getSMBVersion())
+
+            username, domain = (
+                cifutils.splitDomainAndUsername(self.dconf['username'])
+            )
+
+            if domain:
+                options.append('domain=' + domain)
         except:
             util.SMlog("Exception while attempting to append mount options")
             raise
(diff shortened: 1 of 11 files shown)


drivers/cifutils.py
```diff
@@ -0,0 +1,57 @@
...
+import SR, VDI, SRCommand, util
+import nfs
+import os, re
+import xs_errors
+
+
+def getCIFCredentials(dconf, session):
+    credentials={}
+    if dconf.has_key('username') \
+            and (dconf.has_key('cifspassword') or dconf.has_key('cifspassword_secret')):
```
MarkSymsCtx: The SMB SR just uses password/password_secret so this needs to be overriddeble by the caller

drivers/SMBSR.py
```diff
@@ -109,7 +110,7 @@ def mount(self, mountpoint=None):
             raise SMBException("Failed to make directory: code is %d" %
                                 inst.code)
 
-        self.credentials = os.path.join("/tmp", util.gen_uuid())
+        self.credentials = os.path.join("/run/sm", util.gen_uuid())
```
MarkSymsCtx: Should be redundant now.

drivers/util.py
```diff
@@ -130,9 +130,9 @@ def _getDateString():
     return "%s-%s-%s:%s:%s:%s" % \
           (t[0],t[1],t[2],t[3],t[4],t[5])
 
-def doexec(args, inputtext=None):
+def doexec(args, inputtext=None, new_env={}):
     """Execute a subprocess, then return its return code, stdout and stderr"""
-    proc = subprocess.Popen(args,stdin=subprocess.PIPE,stdout=subprocess.PIPE,stderr=subprocess.PIPE,close_fds=True)
+    proc = subprocess.Popen(args,stdin=subprocess.PIPE,stdout=subprocess.PIPE,stderr=subprocess.PIPE,close_fds=True, env=dict(os.environ.items() + new_env.items()))
```
MarkSymsCtx: Just `env=dict(os.environ, new_env)` I think

drivers/cifutils.py
```diff
@@ -0,0 +1,57 @@
...
+            raise xs_errors.XenError('ISOMountFailure', opterr=err_str)
+
+        if dconf.has_key('cifspassword_secret'):
+            password = util.get_secret(session,dconf['cifspassword_secret'])
+        else:
+            password = dconf['cifspassword']
+
+        domain = util.to_plain_string(domain)
+
+        credentials["username"] = util.to_plain_string(username)
```
MarkSymsCtx: as the keys of this will map directly to the environment variables they need to be USER and PASSWD. Not sure how the domain gets handled.

MarkSymsCtx: Looks like domain cannot be passed by environment, which is really irritating. However it's less of a problem to pass domain via the command line option to `mount`.

drivers/SMBSR.py
BenSimsCitrix: Fixed

drivers/util.py
BenSimsCitrix: Fixed

drivers/cifutils.py
BenSimsCitrix: doh! Fixed

tests/test_ISOSR.py
```diff
@@ -7,7 +7,7 @@
 import SR
 import errno
 import testlib 
-
+import pprint
```
MarkSymsCtx: This doesn't appear to be used?

drivers/cifutils.py
```diff
@@ -0,0 +1,74 @@
...
+        if username_overide:
+            uname = username_overide
+        else:
+            uname = dconf['username']
+
+        username, domain = splitDomainAndUsername(uname)
+
+        credentials["USER"] = util.to_plain_string(username)
+
+        if 'cifspassword_secret' in dconf:
```
MarkSymsCtx: This won't work for the SMBSR which just uses `password` & `password_secret` dconf keys

tests/test_ISOSR.py
```diff
@@ -233,6 +274,7 @@ def test_attach_smb_via_xemount_no_version(self, context,
         context.setup_error_codes()
         smbsr = self.create_smbisosr(options='-o username=administrator,password=password')
         smbsr.attach(None)
+        pread.assert_not_called()
```
MarkSymsCtx: This is new in 3.5 and without `autospec` mock will just invent a new `assert_not_called` method on the mock :(

Need to use `self.assertEqual(0, pread.call_count)`.

If possible add the `autospec=True` attribute to the `mock.patch` attribute as well but that may have other issues, it doesn't always play nice with optional parameters.

drivers/cifutils.py
```diff
@@ -0,0 +1,74 @@
...
+        elif len(dom_username) == 2:
+            domain = dom_username[0]
+            username = dom_username[1]
+        else:
+            raise CIFSException("A maximum of 2 tokens are expected "
+                                   "(<domain>\<username>). {} were given."
+                                   .format(len(dom_username)))
+        return username, domain
+
+def getCIFCredentials(dconf, session, username_overide=None):
```
MarkSymsCtx: What is `username_override`? Doesn't seem to be used anywhere?

drivers/SMBSR.py
```diff
@@ -108,27 +108,27 @@ def mount(self, mountpoint=None):
...
+
+    def mount(self, mountpoint=None):
+
+        mountpoint = self.makeMountPoint(mountpoint)
 
-        self.credentials = os.path.join("/tmp", util.gen_uuid())
+        new_env, domain = cifutils.getCIFCredentials(self.dconf, self.session)
 
-        options = self.getMountOptions()
+        options = self.getMountOptions(domain)
```
MarkSymsCtx: I wonder if there is a further (future?) refinement possible here to combine getting options and environment into a single method that avoids passing the domain back from the getCredentials and then passing it into the getMountOptions. Probably too much for this change so maybe file an enhancement ticket for the future?

tests/test_ISOSR.py
BenSimsCitrix: fixed

drivers/cifutils.py
BenSimsCitrix: Ahh now i understand your comment from above. 

Will fix.

drivers/cifutils.py
BenSimsCitrix: This was part of my general misunderstanding, will fix

drivers/SMBSR.py
BenSimsCitrix: I agree.
Ive seen a lot of copy and paste between ISOSR and SMBSR, which can be farmed of into a util file as part of any enhancement.

drivers/cifutils.py
```diff
@@ -0,0 +1,84 @@
...
+
+class CIFSException(Exception):
+    def __init__(self, errstr):
+        self.errstr = errstr
+
+def getDconfPasswordKey(smbsr=False):
+    key_password = 'cifspassword'
+    key_secret = 'cifspassword_secret'
+
+    if smbsr:
```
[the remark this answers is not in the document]
BenSimsCitrix: good idea.

tests/test_ISOSR.py
```diff
@@ -171,58 +221,72 @@ def test_attach_with_smb_version_3(self, context, _checkmount, pread,
...
+                                  '/var/run/sr-mount/asr_uuid', '-o',
+                                  'cache=none,guest,vers=3.0'], True, {})
 
     @testlib.with_context
     @mock.patch('util.gen_uuid')
     @mock.patch('util.makedirs')
     @mock.patch('ISOSR.ISOSR._checkTargetStr')
     @mock.patch('ISOSR.ISOSR._checkmount')
-    def test_attach_smb_via_xemount_version_1(self, context, _checkmount,
+    @mock.patch('util.pread')
```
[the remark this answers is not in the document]
BenSimsCitrix: I did break the tests, but i fixed it :)

drivers/cifutils.py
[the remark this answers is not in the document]
BenSimsCitrix: Fixed.

drivers/cifutils.py
BenSimsCitrix: fixed

drivers/util.py
BenSimsCitrix: That did not work, unfortunately, I had to use .update

drivers/SMBSR.py
BenSimsCitrix: As you found out experimentally, the man page is wrong here.

Looking at the source has the comment
/*
		 * Note that the password will not be retrieved from the
		 * USER env variable (ie user%password form) as there is
		 * already a PASSWD environment varaible
		 */"

I have created the ticket CA-330305 for a general cleanup.

MarkSymsCtx: > As you found out experimentally, the man page is wrong here.
> 
> Looking at the source has the comment
> /*
> * Note that the password will not be retrieved from the
> * USER env variable (ie user%password form) as there is
> * already a PASSWD environment varaible
> */"
> 
> I have created the ticket CA-330305 for a general cleanup.

Indeed, I think the manpage is completely wrong, it doesn't seem to even work with the domain in the user name on the commandline. it probably did work once but the codes been updated without the associated docs.

drivers/ISOSR.py
```diff
@@ -2,13 +2,13 @@
 #
 # Copyright (C) Citrix Systems Inc.
 #
-# This program is free software; you can redistribute it and/or modify
```
MarkSymsCtx: We should probably avoid all the trailing whitespace changes in this commit so that it's clear what is actually changing.

BenSimsCitrix: Sorry slighly over zealous. 
Will fix.

BenSimsCitrix: Fixed

PR #464, opened 2019-11-08
CP-32433: refine the conditions under which mpath count will trigger

MarkSymsCtx:
Squash noisy logs from checking if mpath is working.


MarkSymsCtx:
Unfortunately github can't tell the rules file is a rename and update so that's a bit difficult to review I'm afraid :(

PR #467, opened 2019-11-20
CP-32204: Dynamic limits to leaf coalesce using storage speed estimate.

BenSimsCitrix:
This needs some extra integration testing, as i'm nervous of the effects.

To date this has been through the gcLVM and gcEXT no problems, I have also provoked a leaf coalesce manually and observed the creation of the speed log file, present and correct.

There is also the question of the persistence of the speed log, do we want it to persist after a reboot? Also do we want to clean it up during SR forget?

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -19,6 +19,7 @@
 #
 
 import os
+import os.path
 import sys
 import time
 import signal
@@ -30,6 +31,7 @@ import traceback
 import base64
 import zlib
 import errno
+import stat
 
 import XenAPI
 import util
@@ -73,6 +75,8 @@ DEFAULT_COALESCE_ERR_RATE = 1.0/60
 
 COALESCE_LAST_ERR_TAG = 'last-coalesce-error'
 COALESCE_ERR_RATE_TAG = 'coalesce-error-rate'
+SPEED_LOG_ROOT = "/var/run/{uuid}.speed_log"
+N_RUNNING_AVERAGE = 10
 
 class AbortException(util.SMException):
     pass
@@ -459,6 +463,8 @@ class VDI:
 
     LIVE_LEAF_COALESCE_MAX_SIZE = 20 * 1024 * 1024 # bytes
     LIVE_LEAF_COALESCE_TIMEOUT = 10 # seconds
+    TIMEOUT_SAFETY_MARGIN = 0.5 # extra margin when calculating
+                                # feasibility of leaf coalesce
 
     JRN_RELINK = "relink" # journal entry type for relinking children
     JRN_COALESCE = "coalesce" # to communicate which VDI is being coalesced
@@ -585,11 +591,21 @@ class VDI:
                 not self.hidden and \
                 len(self.children) == 0
 
-    def canLiveCoalesce(self):
+    def canLiveCoalesce(self, speed):
         """Can we stop-and-leaf-coalesce this VDI? The VDI must be
         isLeafCoalesceable() already"""
-        return self.getSizeVHD() <= self.LIVE_LEAF_COALESCE_MAX_SIZE or \
-                self.getConfig(self.DB_LEAFCLSC) == self.LEAFCLSC_FORCE
+        feasibleSize = False
+        allowedDownTime =\
+                self.TIMEOUT_SAFETY_MARGIN * self.LIVE_LEAF_COALESCE_TIMEOUT
+        if speed:
+            feasibleSize =\
+                self.getSizeVHD()/speed < allowedDownTime
+        else:
+            feasibleSize =\
+                self.getSizeVHD() < self.LIVE_LEAF_COALESCE_MAX_SIZE
+
+        return (feasibleSize or
+                self.getConfig(self.DB_LEAFCLSC) == self.LEAFCLSC_FORCE)
 
     def getAllPrunable(self):
         if len(self.children) == 0: # base case
@@ -786,7 +802,12 @@ class VDI:
 
     def _doCoalesceVHD(vdi):
         try:
+
+            startTime = time.time()
+            vhdSize = vdi.getSizeVHD()
             vhdutil.coalesce(vdi.path)
+            endTime = time.time()
+            vdi.sr.recordStorageSpeed(startTime, endTime, vhdSize)
         except util.CommandException, ce:
             # We use try/except for the following piece of code because it runs
             # in a separate process context and errors will not be caught and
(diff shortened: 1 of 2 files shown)


BenSimsCitrix:
As Tim points out we don't need to store the actual values to get the running average, but maybe we want this to persist as a record of the trends/noise in storage speed?

We should also add the speed logs to the bug tool as well?

olivierlambert:
If you like, we can make a test package for some people we know having coalesce issues on current XCP-ng 8.0.

This way, we could validate for them if it's better (or not).

What do you think? To be fair, this kind of improvements will be REALLY great and will be helpful to understand better what's going on. Do you have plan to expose those metrics/info to XAPI?

BenSimsCitrix:
> If you like, we can make a test package for some people we know having coalesce issues on current XCP-ng 8.0.
> 
> This way, we could validate for them if it's better (or not).
> 
That would be great but I think you'd be brave to try this on a customer system yet, let me code up some integration tests to hammer it and then give it to customers. In the mean time if you have seen particularly nasty snapshot chains that would make good test cases let me know.
> What do you think? To be fair, this kind of improvements will be REALLY great and will be helpful >to understand better what's going on. Do you have plan to expose those metrics/info to XAPI?
It would be nice to have this data in XAPI at least for debugging purposes, but first priority is to get the basics working.

TimSmithCtx:
> As Tim points out we don't need to store the actual values to get the running average, but maybe
> we want this to persist as a record of the trends/noise in storage speed?
> 
> We should also add the speed logs to the bug tool as well?

I'm not convinced any logs we're keeping would cover enough time to tell us anything interesting (and if they do we're probably allowing too much old data into the average). I'm also not sure what we'd actually *do* with such data, so I think just keep total/average/count for now.

olivierlambert:
1. Currently, even with some patches that were imported in 7.1 to fix zombie coalesce process. It solves the zombies issues, but we still have uncoalesced chain on the last leaf for various customers (all of them using LVMoISCSI or oHBA, which is BTW by far triggering a lot of support requests vs NFS -none so far-).
2. Agreed, makes sense to start strep by step, just telling that to keep that in mind for the future (ie for the way you design it today), it's a crucial info for backup purpose. For example, in XO, we have a "security" computing if a chain need coalesce before actually doing another snapshot, [see this for ref](https://xen-orchestra.com/blog/xenserver-coalesce-detection-in-xen-orchestra/)). We are raising the level of coalesce to 2 because of the issue in my first point.

Anyway, when you think it's enough solid to make a first test at scale, we have the opportunity to create package in `testing` repo, and a large community to play with it. This way, you can be more confident when deciding to bundle it into CH :+1:

drivers/cleanup.py
```diff
@@ -585,11 +589,19 @@ def isLeafCoalesceable(self):
...
-    def canLiveCoalesce(self):
+    def canLiveCoalesce(self, speed):
         """Can we stop-and-leaf-coalesce this VDI? The VDI must be
         isLeafCoalesceable() already"""
-        return self.getSizeVHD() <= self.LIVE_LEAF_COALESCE_MAX_SIZE or \
-                self.getConfig(self.DB_LEAFCLSC) == self.LEAFCLSC_FORCE
+        feasibleSize = False
+        if speed:
+            feasibleSize =\
+                self.getSizeVHD()/speed < self.LIVE_LEAF_COALESCE_TIMEOUT
```
MarkSymsCtx: I think we should be conservative here and divide the timeout by 2 to make sure that we get done within the time cutoff

drivers/cleanup.py
```diff
@@ -1811,7 +1971,11 @@ def _snapshotCoalesce(self, vdi):
             return False
         Util.log("Coalescing parent %s" % tempSnap)
         util.fistpoint.activate("LVHDRT_coaleaf_delay_2", self.uuid)
+        vhdSize = vdi.getSizeVHD()
+        startTime = time.time()
         self._coalesce(tempSnap)
+        endTime = time.time()
+        self.recordStorageSpeed(startTime, endTime, vhdSize)
```
MarkSymsCtx: We should probably do this calculation in the _coalesce method as that's used by both snapshot and leaf coalesce, both of which are useful to the speed calculation. We should be able to get the size of the thing we're trying to coalesce rather than the VDI as that's the actual amount of data we're going to copy

tests/test_cleanup.py
```diff
@@ -487,3 +507,884 @@ def test_file_vdi_delete(self, mock_lock):
...
         mock_lock.Lock.cleanupAll.assert_called_with(str(vdi_uuid))
+
+    def test_coalesceLeaf(self):
+
+        sr_uuid = uuid4()
+        sr = create_cleanup_sr(uuid=str(sr_uuid))
+        vdi_uuid = uuid4()
+        vdi = cleanup.VDI(sr, str(vdi_uuid), False)
+
+        vdi.canLiveCoalesce = mock.MagicMock(autospec=True)
```
MarkSymsCtx: Do you not have to tell this what the spec is for autospec to work? Or is it magic when mocking a function?

drivers/cleanup.py
BenSimsCitrix: fixed

drivers/cleanup.py
BenSimsCitrix: Dang i thought about just after i pushed. Fixed.

tests/test_cleanup.py
BenSimsCitrix: Your right, the test won't catch someone calling the function with a wrong signature which was my intention. Annoying that python let me do this.

Mocking the function with a patch and autospec=true does catch it, 

I will refactor the tests to use patch.

BenSimsCitrix: Test refactoring finally done.

tests/test_cleanup.py
```diff
@@ -487,3 +507,810 @@ def test_file_vdi_delete(self, mock_lock):
...
+        mock_vdi.canLiveCoalesce.return_value = False
+        mock_srSnapshotCoalesce.return_value = False
+        mock_srLeafCoalesce.return_value = False
+        sr_uuid = uuid4()
+        sr = create_cleanup_sr(uuid=str(sr_uuid))
+        vdi_uuid = uuid4()
+        vdi = cleanup.VDI(sr, str(vdi_uuid), False)
+
+        res = sr._coalesceLeaf(vdi)
+        self.assertEquals(res, False)
```
MarkSymsCtx: `self.assertFalse`?

tests/test_cleanup.py
```diff
@@ -487,3 +507,810 @@ def test_file_vdi_delete(self, mock_lock):
...
+
+        sr_uuid = uuid4()
+        sr = create_cleanup_sr(uuid=str(sr_uuid))
+        vdi_uuid = uuid4()
+        vdi = cleanup.VDI(sr, str(vdi_uuid), False)
+
+        with self.assertRaises(util.SMException) as exc:
+            res = sr._coalesceLeaf(vdi)
+
+        mock_exception_init.assert_called_with("VDI {uuid} could"
```
MarkSymsCtx: You should be able to test the content of `exc.message`, not sure if that's any easier.

tests/test_cleanup.py
```diff
@@ -487,3 +507,810 @@ def test_file_vdi_delete(self, mock_lock):
...
+
+        res = sr._coalesceLeaf(vdi)
+
+        self.assertEqual(res, "This is a Test")
+        self.assertEqual(4, mock_vdiLiveCoalesce.call_count)
+        self.assertEqual(3, mock_snapshotCoalesce.call_count)
+        self.assertEqual(6, mock_vhdSize.call_count)
+
+    @mock.patch('cleanup.Util.log')
+    def test_findLeafCoalesceable_forbiden1(self, mock_log):
```
MarkSymsCtx: forbidden?

tests/test_cleanup.py
```diff
@@ -487,3 +507,810 @@ def test_file_vdi_delete(self, mock_lock):
...
+    @mock.patch('cleanup.VDI.getConfig', autospec=True)
+    def test_gather_candidates_leaf_not_coalescable(self, mock_getConfig,
+                                                    mock_isLeafCoalesceable,
+                                                    mock_leafCoalesceForbidden
+                                                    ):
+
+        """ The bad vdi returns false for isLeafCoalesceable and is not
+            added to the list.
+        """
+        self.gather_candidates(mock_getConfig,
```
MarkSymsCtx: nothing to check in these tests? How do we know they've done the right thing?

tests/test_cleanup.py
```diff
@@ -487,3 +507,810 @@ def test_file_vdi_delete(self, mock_lock):
...
+        with self.assertRaises(Exception):
+            sr.writeSpeedToFile(1.8)
+        mock_open.assert_called_with(expectedPath, 'r+')
+        self.assertEqual(FakeFile.close.call_count, 1)
+        FakeFile.readlines.side_effect = None
+
+        # File does exist
+        readLines = ["1.9\n", "2.1\n", "3\n"]
+        writeLines = ["1.9\n", "2.1\n", "3\n", "1.8\n"]
+        self.writeSpeedFile(sr, 1.8, mock_isFile, True, mock_open, mock_chmod,
```
MarkSymsCtx: What are these tests achieving, just code coverage and no validation of functionality?

tests/test_cleanup.py
BenSimsCitrix: :)

tests/test_cleanup.py
BenSimsCitrix: fixed

tests/test_cleanup.py
BenSimsCitrix: saves a mock, fixed.

drivers/cleanup.py
```diff
@@ -1555,6 +1577,26 @@ def findLeafCoalesceable(self):
 
         return None
 
+    def gatherCandidates(self, candidates):
```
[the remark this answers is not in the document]
BenSimsCitrix: Done

tests/test_cleanup.py
BenSimsCitrix: fixed

tests/test_cleanup.py
BenSimsCitrix: I refactored the actual asserts into gather_candidates, The asserts definitely fire in case of an error.

Perhaps there is something reason not to put the assert in sub function, rather than a test_function??

tests/test_cleanup.py
BenSimsCitrix: I have refactored the common assert code into writeSpeedFile.

In other test frameworks this pattern is ok, but perhaps there is something wrong with doing this in python unittest.

MarkSymsCtx: No, that's fine, I just hadn't seen that was what was happening :)

tests/test_cleanup.py
```diff
@@ -487,3 +507,808 @@ def test_file_vdi_delete(self, mock_lock):
...
+
+        mock_vhdSize.side_effect = iter([1024, 4096, 4096, 8000, 8000, 16000])
+
+        sr._snapshotCoalesce = mock.MagicMock(autospec=True)
+        sr._snapshotCoalesce.return_value = True
+
+        with self.assertRaises(util.SMException) as exc:
+            res = sr._coalesceLeaf(vdi)
+
+        mock_exceptionInit.assert_called_with("VDI {uuid} could"
```
MarkSymsCtx: This should be the same as the method above?

BenSimsCitrix: dang

BenSimsCitrix: fixed

olivierlambert:
Nice :+1: Do you consider this "testable"? We have some "cutting edge" users who might want to play with it :smile:

BenSimsCitrix:
> Nice +1 Do you consider this "testable"? We have some "cutting edge" users who might want to play with it smile

Feel free to beta test, with the caveat that this has not gone through the usual stress testing of an official Citrix hypervisor release yet.

olivierlambert:
@BenSimsCitrix that's fine to me :+1:  Hopefully, we can bring "field testing feedback" on our side for some people who want to experiment. This will be mutually beneficial for both of us :)

We'll keep you posted. Pinging @stormi about packaging this version and put it in our testing repo.

BenSimsCitrix:
This is ready to merge

PR #468, opened 2019-12-01
Replacing ls command with "find" enables a depth parameter to be set 

imakiro:
Created to solve : https://github.com/xcp-ng/xcp/issues/313
it's tested to not be impacting other commands on my XCP-NG and shouldn't impacting other commands.
The project maintainers might want to set a more conservative default, but I haven't see a major impact on our 20.000 files smb share.


imakiro:
I can't seem to edit my first commit to include a signoff ...

PR #469, opened 2019-12-01
Enable recursive ISOSR scanning

imakiro:
This change from ls to find enable a simple recursion control.


MarkSymsCtx:
I think there are good reasons why we would not want this to be enabled by default given the significant change in behaviour.

imakiro:
> 
> 
> I think there are good reasons why we would not want this to be enabled by default given the significant change in behaviour.

This seemed like a good idea, do you mind explaining which element is impacted significantly ? using a sane default of 1 would simply present the same behaviour as right now.

Nick-085:
Is this something that can be reviewed again? I'm trying not to present every path on my TrueNAS to the network. I already have too many.

olivierlambert:
@Nick-085 have you tried the modifications? It might be easy to test, and having more feedback might help the upstream to be more confident on merging this at some point :)

PR #473, opened 2019-12-17
CA-332890: trigger mpath count when multipath dm device added/removed

MarkSymsCtx:


udev/65-multipath.rules
```diff
@@ -8,6 +8,9 @@ ACTION=="remove", ENV{DM_MULTIPATH_DEVICE_PATH}=="1",  GOTO="count_mpath"
 IMPORT{db}="DM_ACTION"
 ACTION=="change", ENV{DM_ACTION}=="PATH_REINSTATED", GOTO="count_mpath"
 ACTION=="change", ENV{DM_ACTION}=="PATH_FAILED", GOTO="count_mpath"
+ACTION=="add", PROGRAM=="/bin/bash -c '/sbin/dmsetup table -j %M -m %m | /bin/grep multipath'", ENV{MULTIPATH}="1"
```
MarkSymsCtx: Not sure whether this variable should get a product specific prefix or not.

udev/65-multipath.rules
```diff
@@ -8,6 +8,9 @@ ACTION=="remove", ENV{DM_MULTIPATH_DEVICE_PATH}=="1",  GOTO="count_mpath"
...
+IMPORT{db}="MULTIPATH"
+ACTION=="add|remove", ENV{MULTIPATH}=="1", GOTO="count_mpath"
```
BenSimsCitrix: Can lines 6 and 7  be removed now because of this?

BenSimsCitrix: Ah sorry i see, that is matching is against another ENV variable.

MarkSymsCtx: Yes, add/remove DM_MULTPATH_DEVICE_PATH is adding a sub device of the multipath

udev/65-multipath.rules
BenSimsCitrix: It might be wise to avoid any name collisions.

BenSimsCitrix: Do we care about the $? return of this PROGRAM? according to udev man page the stdout goes to RESULT but the key gets set to true if the program succeeds (I assume this means  $? == 0)  which will always be true as running "bash -c" swallows the $?

I think running under bash -ce will propagate the $?

PR #474, opened 2019-12-18
CA-332806 - Lock speedfile and use atomic write to prevent corruption…

BenSimsCitrix:
This has got through 

storageGCfile 2756961 
storageGClvm 2756960

and importantly
storageLeafCoalesceLVM 2756959

I'm running it through a BST (120598)  when that is clear, I will remove the do not merge flag

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -75,7 +75,9 @@ DEFAULT_COALESCE_ERR_RATE = 1.0/60
 
 COALESCE_LAST_ERR_TAG = 'last-coalesce-error'
 COALESCE_ERR_RATE_TAG = 'coalesce-error-rate'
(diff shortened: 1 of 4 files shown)


drivers/cleanup.py
```diff
@@ -189,8 +191,9 @@ def runAbortable(func, ret, ns, abortTest, pollInterval, timeOut):
                 else:
                     resultFlag.set("failure")
             except Exception, e:
-                resultFlag.set("failure")
                 Util.log("Child process failed with : (%s)" % e)
+                resultFlag.set("failure")
+                Util.logException("This exception has occured")
```
BenSimsCitrix: This seems like useful debug, If an exception occurs in runAbortable we don't see a stack trace in the log.

BenSimsCitrix:
120598 completed without incident.

PR #482, opened 2020-02-12
CP-31856 - Return nbd socket as parameter to XAPI

BenSimsCitrix:
This has had minimal testing, (VM boot, reboot)

Travis does not appear to like this yet, Yet the the unit tests are fine locally??

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -526,6 +526,17 @@ def mkdirs(path, mode=0777):
             if e.errno != errno.EEXIST:
                 raise
 
+def rmdirs(path):
+    # lexist returns true even if link is broken
+    if os.path.lexists(path):
+        parent, subdir = os.path.split(path)
+        assert parent != path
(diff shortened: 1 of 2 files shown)


BenSimsCitrix:
Unit tests passing now. 

Going to do a koji build and run it through BVT

BenSimsCitrix:
koji has finished, storage transition building.

drivers/blktap2.py
```diff
@@ -1652,6 +1680,7 @@ def _activate_locked(self, sr_uuid, vdi_uuid, options):
 
         # Link result to backend/
         self.BackendLink.from_uuid(sr_uuid, vdi_uuid).mklink(dev_path)
+ 	self.linkNBD(sr_uuid, vdi_uuid)
```
MarkSymsCtx: Looks to be some duplicated code here with what's on line ~ 1553. A cleanup in a subsequent commit would be good.

PR #486, opened 2020-03-03
CA-335351: ensure async tasks are cleaned up

MarkSymsCtx:


drivers/cleanup.py
```diff
@@ -405,20 +405,28 @@ def srUpdate(self):
...
+                        cancelTask = False
+                        Util.log("Task cancelled")
+                    except:
+                        pass
+                    return
+                time.sleep(1)
+        finally:
+            if cancelTask:
+                self.session.xenapi.task.cancel(task)
+            self.session.xenapi.task.destroy(task)
```
BenSimsCitrix: Does a task need to be canceled to be destroyed?

MarkSymsCtx: That was the advice I had from the toolstack folks. It won't hurt anyway to cancel - destroy.

PR #487, opened 2020-03-04
Changed default NFS mount option for non ISO SRs to 'hard'

ezaton:
Modified soft_mount function to call a new function 'mount' with the 'soft' argument. Otherwise - the function 'mount' works exactly like the past soft_mount, except that it gets an extra parameter - 'recovery_mode', which, by default, is set to hard - except when called from soft_mount.
I have changed the files calling this function (as nfs.mount instead of nfs.soft_mount) - NFSSR.py, test_NFSSR.py and test_nfs.py - to call a 'hard' mount by default.

Some extended explanation:
NFS soft mount means that a short disconnection of NFS SR would be visible to the virtual machines. This is very good for storage recovery and handling of stale NFS mounts - because the application (in this case - the VM) get an IO error, and can abort. While this can be desirable for any regular NFS mount client, this is far from optimal for NFS shares used for VMs. It has been bugging me for a long while now, and I usually handle it by hacking the file nfs.py manually, enforcing hard mount. Short NFS disconnections can happen when an NFS cluster performs failover/takeover, when there is a small glitch in the network. With the current NFS soft mount option (which cannot be overridden even when using NFS SR arguments!) the NFS break "climbs" up to the virtual machines. In Windows a message saying "Delayed write failed" in the tray. In Linux it would, usually, cause the VM to switch filesystems to read-only mode, which would, in turn, require a reboot to resolve (emphasis on root filesystem).
The desired behaviour is 'hard' nfs recovery mount. If the NFS share becomes inaccessible, freeze all IO meant to get to it, until it becomes available again. It will still maintain crash consistency, which is imperative for VMs, and it might make life harder when you want to umount an SR (stale NFS are no fun), but it will reduce the impact of any network hang or cluster failover to zero.
I have been testing the manual hack on several XenServers and XCP-NG systems for a while now, and following a conversation I've had here, it is a good idea to bring this change to the mainstream. I have, purposely, left ISO mounts 'soft', because ISO shares can (and sometimes are), based on my experience, be placed on a more mobile systems (desktops, laptops even) - so I would not want failed mount there to block the system.

On another note - VMware mounts NFS hard. Always has. It works flawlessly there.

### DIFF
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -2,13 +2,13 @@
 #
 # Copyright (C) Citrix Systems Inc.
 #
-# This program is free software; you can redistribute it and/or modify 
-# it under the terms of the GNU Lesser General Public License as published 
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
 # by the Free Software Foundation; version 2.1 only.
 #
-# This program is distributed in the hope that it will be useful, 
-# but WITHOUT ANY WARRANTY; without even the implied warranty of 
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 # GNU Lesser General Public License for more details.
 #
 # You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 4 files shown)


MarkSymsCtx:
I don't think this will be accepted into the mainstream. The XenServer default was deliberately changed from hard to soft a long time back.

stormi:
Do the reasons that applied back then still apply? What were they?

MarkSymsCtx:
> Do the reasons that applied back then still apply? What were they?

Not being able to reboot hosts if the NFS storage is unreachable is one.

stormi:
An update: I have personnally no preference, I'm just looking for the best solution. Having other major actors mount NFS as hard is an interesting pointer. The discussion between users (and their testing reports) at https://github.com/xcp-ng/xcp/issues/334 is rich and may help deciding.

I think there's a possibility that the drawbacks of hard mounts can be workarounded and be of less severity than sending IO errors to all running VMs.

PR #492, opened 2020-03-23
CA-331454: do not post attach scan ISO SRs

MarkSymsCtx:


MarkSymsCtx:
> There is loads of copy and paste between ISOSR and SMBSR, might be worth checking the same problem won't pop up there.

The problem here was the tools pseudo SR which is an ISO SR but doesn't have any PBDs when being attached and so the scan fails.

PR #496, opened 2020-04-14
CP-33292: add read caching capability flags

MarkSymsCtx:


edwintorok:
How about CIFS SR? XenRT thinks it would support it.

edwintorok:
https://github.com/xapi-project/xen-api/pull/4076 toolstack part

MarkSymsCtx:
> How about CIFS (ISO?) SR? XenRT thinks it would support it.

ISOs are opened O_DIRECT by tapdisk so not cached.

PR #501, opened 2020-05-19
CA-332978: fix SR scan with failed multipath path

MarkSymsCtx:


drivers/mpath_dmp.py
```diff
@@ -234,12 +234,18 @@ def _is_valid_multipath_device(sid):
...
     if not _is_valid_multipath_device(sid):
         return
-    util.retry(lambda: util.pread2(['/usr/sbin/multipath', '-r', sid]), maxretry = 3,
-                           period = 4)
     path = os.path.join(DEVMAPPERPATH, sid)
-    util.wait_for_path(path, 10)
+    # If the mapper path doesn't exist force a reload in multipath
     if not os.path.exists(path):
-        raise xs_errors.XenError('DMP failed to activate mapper path')
+        util.retry(lambda: util.pread2(
```
BenSimsCitrix, line 240: Just educating myself about the "util" functions. Don't we need to use "pread" instead of pread2 and give an expected return code, or it won't raise an exception and util.retry wont try again as it is waiting to catch an exception?

I see there are several example of this pattern in the file, so maybe i'm wrong ...

MarkSymsCtx: That's possible but if so it was wrong before as this has just moved from line 237 and inside an if.

BenSimsCitrix: Ahh the default return code is 0.

So as long as we expect that for success we are ok :)

MarkSymsCtx: I suspect it's also entirely unnecessary to be in a retry at all to be honest. I'm not aware of any way in which `multipath -r` can error.

MarkSymsCtx: Actually pread2 calls pread with the default rc=0

BenSimsCitrix: Ive been educated.

PR #506, opened 2020-06-29
Stage 1 futurize on SM

MarkSymsCtx:
Mechanical code update using

futurize --stage1

https://python-future.org/futurize.html

### DIFF
--- a/tests/test_mpathcount.py
+++ b/tests/test_mpathcount.py
@@ -1,6 +1,7 @@
 """
 Unit tests for mpathcount
 """
+from __future__ import print_function
 import errno
 import os
 import unittest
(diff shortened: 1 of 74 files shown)


TimSmithCtx (approved):

tests/test_mpathcount.py
```diff
@@ -1,6 +1,7 @@
 """
 Unit tests for mpathcount
 """
+from __future__ import print_function
```
BenSimsCitrix, line 4: Doesn't do any harm but any idea why its import future here and not used it.

BenSimsCitrix: Ahh there a print on line 120, that must be why.

BenSimsCitrix (commented):
LGTM

BenSimsCitrix (approved):
LGTM

PR #508, opened 2020-07-14
CP-32988: Move tapdisks into configured cgroup

BenSimsCitrix:


BenSimsCitrix:
Whether we use 25% is upto the further testing.

pagecache is counted towards the memory cgroup as this chap from linkedin found out.

https://engineering.linkedin.com/blog/2016/08/don_t-let-linux-control-groups-uncontrolled

blktap/blktap.cgroups
```diff
@@ -0,0 +1,7 @@
+{
+   "memory": {
+     "MaxPercentageRam": 25.0
+   },
+   "blkio":{
```
MarkSymsCtx, line 5: Does this override the blkio setting in blktap itself? Blktap configures vm.slice for blkio and that is configured
```
[Unit]
Description=VM blkio slice

[Slice]
BlockIOWeight=100

[Install]
WantedBy=multi-user.target
```

drivers/blktap2.py
```diff
@@ -60,7 +60,76 @@
...
+    if util.pathexists(CONF_FILE):
+        with open(CONF_FILE) as json_file:
+            data = json.load(json_file)
+            cgroup_create(data)
+            _cgroups = list(data.keys())
+            return _cgroups
+    return None
+
+
+def cgroup_move(pid, _cgroups):
```
MarkSymsCtx, line 126: I don't understand why we're not using the existing functionality in blktap for this? 
Calling `tap-ctl spawn -s <slice name>` should already place the spawned process into the specified cgroup.

BenSimsCitrix: We could use it for the first cgroup, but it would need to be extended to take a list of cgroups.

Philosophically why do we prefer an application moving itself into a cgroup. This feels like something that should be done a layer above?

blktap/blktap.cgroups
BenSimsCitrix: Ahh Things seem more complicated for some of the cgroups.
blkio has a vm.slice but memory does not.
I could change the behaviour so that we create a tapdisk cgroup underneath any preexisting vm.slice? Or we could create vm.slice for every cgroup ? What do you think?

blktap/blktap.cgroups
```diff
@@ -0,0 +1,7 @@
+{
+   "memory": {
+     "MaxPercentageRam": 25.0
```
edwintorok, line 3: Also we might want to set `memory.swampiness` to some low value (e.g. `1`, I wouldn't set it to 0 because that'd invoke the OOM killer, swapping might still be preferable to OOM killing).

We probably need to find a good value through experimentation, or dynamically adjust based on memory pressure (not sure if that is possible). 25% seems quite low, after all if you give Dom0 extra RAM it might be because you want more available for read caching. An idle system seems to have ~85% free Dom0 RAM (including pagecache), OTOH one with lots of qemus might have less. 75% might be a better starting point.

One possible experiment could be:
* open a large file over NFS RO, and read from it in parallel from multiple processes (e.g. with appropriate `fio` flags in a loop and lots of processes in the hundreds, simulating VMs accessing a base disk)
* in parallel that do something that allocates/frees memory with periodic pauses, e.g. several VDI exports, and parallel VM writes (these would allocate tapdisk buffers, and free them again when idle)

I would expect there is MaxPercentageRam below 100 where it all works well, and above which you start seeing pauses/latency spikes. Then we can pick a value below that high threshold as default (and ideally add a xenrt test, but for now I'd be happy with just manual confirmation).

edwintorok: Also it might be useful to be able to override this at runtime for easier experimentation (and also to solve customer issues without doing unsupported modifications to Dom0 fs). E.g. VBD polling thresholds are configurable through XAPI.
Maybe we could use a field in `sm_config` to configure this (and of course have a suitable default if that is not set)

olivierlambert:
Hi guys!

We are here if you need to do some testing :) Also, can you develop in few words what's the motivation for this PR? I mean the goal of move tapdisks into configured cgroup?

BenSimsCitrix:
So pretty much trashed my old implementation after educating myself about the tools available in centos for cgroup manipulation.

We now use cgclassify to move the tapdisk into the correct cgroups dependant on the rules in cgrules.conf. This is what we used to do 3 years ago but back then we were using cgexec which sometimes failed because the tap-ctl process is short-lived.

blktap/blktap.cgroups
BenSimsCitrix: Hi Edvin,  using the new implementation, we just need to edit the vm-blktap.slice file and then do a "systemctl enable vm-blktap.slice" and a "systemctl restart vm-blktap.slice". Of course this would change the behaviour of all tapdisks.

If you want to configure individual tapdisks that would be way more complex.

BenSimsCitrix: Once people think this approach is correct, I can start to run some experiments to determine the correct ram percentage.

blktap/blktap.cgroups
BenSimsCitrix: The cgrules now comply with the previous behaviour in addition to moving tapdisks to the Memory cgroup.

drivers/blktap2.py
BenSimsCitrix: There is some history to this issue in CA-225067. We used to use the cgrules.conf by using cgexec but this was found to be racy due the shortlivedness of tap-ctl. So the moving the process became the responsibility of tap-ctl which has future-proofing issues if we ever move to v2 cgroups.

I don't think there is any issue in getting SM to move the process to the vm.slice and vm-blktap.slice as the tapdisk is not connected to anything at this point and is freshly spawned.

BenSimsCitrix:
> Hi guys!
> 
> We are here if you need to do some testing :) Also, can you develop in few words what's the motivation for this PR? I mean the goal of move tapdisks into configured cgroup?

We have found for some customers tapdisk is causing memory starvation in other areas of Dom0 and affecting performance.

olivierlambert:
In those cases, we are increasing dom0 memory, but indeed, it might happen. Anyway, if there's something you consider "testable", we'll be happy to push it into our testing repo so we can submit it to the community :+1:

PR #524, opened 2020-11-13
CP-35367: report GC progress by task

MarkSymsCtx:

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -286,6 +286,8 @@ class XAPI:
         self.srRecord = self.session.xenapi.SR.get_record(self._srRef)
         self.hostUuid = util.get_this_host()
         self._hostRef = self.session.xenapi.host.get_by_uuid(self.hostUuid)
+        self.task = None
+        self.task_progress = {"coalescable": 0, "done": 0}
 
     def __del__(self):
         if self.sessionPrivate:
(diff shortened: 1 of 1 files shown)


olivierlambert:
If it's what I understand, well, that's a very nice feature! Thanks a lot :+1: 

About the coalesce progress exposed, is it a progress toward the total amount of VDI to coalesce, or a progress in data to coalesce per VDI?

MarkSymsCtx:
> If it's what I understand, well, that's a very nice feature! Thanks a lot 👍
> 
> About the coalesce progress exposed, is it a progress toward the total amount of VDI to coalesce, or a progress in data to coalesce per VDI?

It's progress on the number of VHD pairs which means it's a bit non-linear as it takes no account of the size of each VHD to coalesce, partly because it doesn't fully have all that information until it comes to actually do that pair. The total number can increase whilst the GC is running if additionally snapshots are deleted and so the progress could potentially go backwards as well.

olivierlambert:
Okay. Still, it's very interesting and useful. For now, we couldn't know from XO/XAPI perspective if a GC was currently running or not (but we managed to guess when it should be triggered, and prevented doing more snapshot before the chain was fully coalesced see https://xen-orchestra.com/blog/xenserver-coalesce-detection-in-xen-orchestra/).

With this, we can also add that a GC is currently running on the SR (or not), so we can display that information in the "Advanced" view of the appropriate SR. We can even probably extrapolate at some point and report problems because a GC is never done after a while, despite the fact the chain should be coalesced.

Again, very good idea and feature :+1:

TimSmithCtx (approved):

MarkSymsCtx:
> With this, we can also add that a GC is currently running on the SR (or not), so we can display that information in the "Advanced" view of the appropriate SR. We can even probably extrapolate at some point and report problems because a GC is never done after a while, despite the fact the chain should be coalesced.

Hmmmm, I wonder what my motivation for the change was? ;)

BenSimsCitrix (approved):
LGTM

PR #526, opened 2020-11-23
CA-349188: only the initiatorname is critical

MarkSymsCtx:
The installer might not set alias and so iSCSIBoot can fail.


TimSmithCtx (approved):

BenSimsCitrix (approved):
LGTM

PR #535, opened 2021-01-27
XSI-915: Improve performance of LVHDoHBA operations by

BenSimsCitrix:
This commit increases performance by.

       1) Reducing expensive calls to _pathfresh
	2) Use multipath with option "-l" and then falling back to using the more expensive "-ll"
	   only if first operation it fails.
	3) Add missing vdi_epoch operations to vdi operation list

This has been through a BST.

### DIFF
--- a/drivers/LVHDoHBASR.py
+++ b/drivers/LVHDoHBASR.py
@@ -107,14 +107,13 @@ class LVHDoHBASR(LVHDSR.LVHDSR):
                 raise xs_errors.XenError('ConfigSCSIid')
 
         self.SCSIid = self.dconf['SCSIid']
-        self._pathrefresh(LVHDoHBASR, load=False)
         LVHDSR.LVHDSR.load(self, sr_uuid)
 
     def create(self, sr_uuid, size):
         self.hbasr.attach(sr_uuid)
         if self.mpath == "true":
             self.mpathmodule.refresh(self.SCSIid, 0)
-        self._pathrefresh(LVHDoHBASR)
+        self._pathrefresh()
         try:
             LVHDSR.LVHDSR.create(self, sr_uuid, size)
         finally:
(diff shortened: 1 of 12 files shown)


drivers/LVHDoHBASR.py
```diff
@@ -114,7 +114,7 @@ def create(self, sr_uuid, size):
         self.hbasr.attach(sr_uuid)
         if self.mpath == "true":
             self.mpathmodule.refresh(self.SCSIid, 0)
-        self._pathrefresh(LVHDoHBASR)
+        self.load(sr_uuid)
```
MarkSymsCtx, line 117: Not required. This is called from the superclass __init__

drivers/LVHDoHBASR.py
```diff
@@ -132,7 +132,7 @@ def attach(self, sr_uuid):
             for file in os.listdir(path):
                 self.block_setscheduler('%s/%s' % (path, file))
 
-        self._pathrefresh(LVHDoHBASR)
+        self.load(sr_uuid)
```
MarkSymsCtx, line 135: Ditto

drivers/LVHDoHBASR.py
```diff
@@ -150,7 +150,7 @@ def scan(self, sr_uuid):
             if not os.path.exists(self.dconf['device']):
                 util.SMlog("@@@@@ path does not exists")
                 self.mpathmodule.refresh(self.SCSIid, 0)
-                self._pathrefresh(LVHDoHBASR)
+                self.load(sr_uuid)
```
MarkSymsCtx, line 153: Ditto

MarkSymsCtx (changes requested):
Remove the extraneous calls to load

BenSimsCitrix:
This has got through another BST.

BenSimsCitrix:
Pesky "iter" bug got me again this does manifest locally when i run the tests hence why i keep getting bitten, Tests are passing now

tests/test_SR.py
```diff
@@ -0,0 +1,39 @@
...
+    def tearDown(self):
+        pass
+
+    def create_SR(self, cmd, dconf):
+        srcmd = mock.Mock()
+        srcmd.dconf = dconf
+        srcmd.params = {'command': cmd}
+        return SR.SR(srcmd, "some SR UUID")
+
+    def test_checkroot_no_device(self):
```
MarkSymsCtx, line 22: Do these tests need to be asserting anything after the calls to check they functioned correctly?

MarkSymsCtx (approved):

TimSmithCtx (approved):

PR #538, opened 2021-01-28
Add github actions

MarkSymsCtx:


TimSmithCtx (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 2.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
```
edwintorok, line 18: Be careful, the latest version of pip has already dropped Python 2.7 support https://pip.pypa.io/en/stable/news/#id1 (hopefully pypi is smart enough to offer you the last version that still supports it, if not you might need a version constraint)

edwintorok (approved):

.github/workflows/main.yml
MarkSymsCtx: Yep, we'll cross that when it causes trouble. Maybe we'll get transitioned to Python3 before that happens :)

PR #540, opened 2021-01-29
Revert "CA-350871: Remove _lvmBugCleanup, superseded by locks"

MarkSymsCtx:
This reverts commit 87690110ee8f943083c04a51a611efa1b58a6c84.

Even with locks around each call to LVM commands we still get
failures on `lvchange -ay` when called immediately after an
`lvchange -an` on the same volume (as used in the remote host
refresh operation).


TimSmithCtx (approved):

BenSimsCitrix (approved):
LGTM

PR #544, opened 2021-02-04
Additional SM Driver for MooseFS

xandrus:
Hi,

I would like to add an additional SM driver for MooseFS distributed file system.

Thanks!

### DIFF
--- a/drivers/MooseFSSR.py
+++ b/drivers/MooseFSSR.py
@@ -1,0 +1,296 @@
+#!/usr/bin/env python
+#
+# Original work copyright (C) Citrix systems
+# Modified work copyright (C) Tappest sp. z o.o., Vates SAS and XCP-ng community
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; version 2.1 only.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+#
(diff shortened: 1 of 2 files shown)


edwintorok:
Thanks, I think we'd prefer new storage drivers to use SMAPIv3 instead of SMAPIv1 though.
See https://xapi-project.github.io/xapi-storage/#introduction https://github.com/xapi-project/xapi-storage/tree/master/python/examples

SMAPIv1 is difficult to maintain, and has some architectural issues that were corrected in SMAPIv3. Also SMAPIv3 should have better documentation on the semantics of each API call that a storage driver has to implement.

MarkSymsCtx:
Thanks for this, unfortunately we are no longer accepting new SRs
using v1 of the SM API, it is deprecated in preference to the v3
(SMAPIv3) API. See here -
https://xapi-project.github.io/xapi-storage/#introduction with simple
example implementations here -
https://github.com/xapi-project/xapi-storage/tree/master/python/examples.

The new API enforces a cleaner separation between the control plane
and the implementation of the storage management which improves
reliability and facilitates independent testing of the SR
implementation without any specific control plane implementation.

olivierlambert:
I understand the answer, but SMAPIv3 doesn't have:

* any migration mechanism from SMAPIv1
* no VHD nor QCOW2 export format (hard to backup, and NBD implem in XAPI is incomplete)
* still perf issues
* a complete lack of example integration outside a non-documented proprietary Citrix driver

So I agree new drivers should transition to SMAPIv3, but it's not production ready yet for the community to participate. My 2 cents.

xandrus:
Thank you for all the information.

Thanks!

MarkSymsCtx:
SMAPIv3 also allows you to use the full power of the storage rather than being tied to a lowest common denominator of VHD that SMAPIv1 forces on you.

In the case here of MooseFS you don't need, and therefore probably shouldn't use, and form of image format and can use files on the filesystem as direct raw virtual disk files. Snapshot and Cloning can be performed using `mfsmakesnapshot` to perform a server side offloaded snasphot operation. No requirement for any SR level garbage collection and no indirection via a performance bottlenecking image format process.

olivierlambert:
Yes, that's a great thing (and also SMAPIv3 code is far better shape).

However, how would you backup this at the hypervisor level? Exporting the raw content directly?

PR #547, opened 2021-02-11
Support IPv6 in NFS

benjamreis:
Signed-off-by: BenjiReis <benjamin.reis@vates.fr>

### DIFF
--- a/drivers/NFSSR.py
+++ b/drivers/NFSSR.py
@@ -93,9 +93,7 @@ class NFSSR(FileSR.FileSR):
         self.path = os.path.join(SR.MOUNT_BASE, sr_uuid)
 
         # Handle optional dconf attributes
-        self.transport = DEFAULT_TRANSPORT
-        if 'useUDP' in self.dconf and self.dconf['useUDP'] == 'true':
-            self.transport = "udp"
+        self.set_transport()
         self.nfsversion = nfs.validate_nfsversion(self.dconf.get('nfsversion'))
         if 'options' in self.dconf:
             self.options = self.dconf['options']
@@ -256,6 +254,14 @@ class NFSSR(FileSR.FileSR):
         util.SMlog("scanning2 (target=%s)" % target)
         dom = nfs.scan_exports(target)
         print(dom.toprettyxml(), file=sys.stderr)
+
+    def set_transport(self):
+        self.transport = DEFAULT_TRANSPORT
+        use_ipv6 = util.get_ip_address_family(self.remoteserver) == 6
+        if use_ipv6:
+            self.transport = 'tcp6'
+        if 'useUDP' in self.dconf and self.dconf['useUDP'] == 'true':
+            self.transport = 'udp6' if use_ipv6 else 'udp'
 
 
 class NFSFileVDI(FileSR.FileVDI):
--- a/drivers/nfs.py
+++ b/drivers/nfs.py
@@ -161,6 +161,8 @@ def soft_mount(mountpoint, remoteserver, remotepath, transport, useroptions='',
         options += ",%s" % useroptions
 
     try:
+        if transport in ['tcp6', 'udp6']:
+            remoteserver = '[' + remoteserver + ']'
         util.ioretry(lambda:
                      util.pread([mountcommand, "%s:%s"
                                  % (remoteserver, remotepath),
@@ -261,12 +263,13 @@ def get_supported_nfs_versions(server):
     valid_versions = set(['3', '4'])
     cv = set()
     try:
-        ns = util.pread2([RPCINFO_BIN, "-p", "%s" % server])
+        ns = util.pread2([RPCINFO_BIN, "-s", "%s" % server])
         ns = ns.split("\n")
         for i in range(len(ns)):
             if ns[i].find("nfs") > 0:
-                cvi = ns[i].split()[1]
-                cv.add(cvi)
+                cvi = ns[i].split()[1].split(",")
+                for j in range(len(cvi)):
+                    cv.add(cvi[j])
         return list(cv & valid_versions)
     except:
         util.SMlog("Unable to obtain list of valid nfs versions")
--- a/tests/test_NFSSR.py
+++ b/tests/test_NFSSR.py
@@ -89,3 +89,8 @@ class TestNFSSR(unittest.TestCase):
                                            timeout=100,
                                            nfsversion='aNfsversionChanged',
                                            retrans=3)
+
+    @mock.patch('NFSSR.Lock', autospec=True)
+    def test_load_ipv6(self, mock_lock):
+        nfssr = self.create_nfssr(server='::1')
+        self.assertEqual(nfssr.transport, 'tcp6')
--- a/tests/test_nfs.py
+++ b/tests/test_nfs.py
@@ -12,13 +12,13 @@ class Test_nfs(unittest.TestCase):
     def test_check_server_tcp(self, pread):
         nfs.check_server_tcp('aServer')
 
-        pread.assert_called_once_with(['/usr/sbin/rpcinfo', '-p', 'aServer'], quiet=False)
+        pread.assert_called_once_with(['/usr/sbin/rpcinfo', '-s', 'aServer'], quiet=False)
 
     @mock.patch('util.pread', autospec=True)
     def test_check_server_tcp_nfsversion(self, pread):
         nfs.check_server_tcp('aServer', 'aNfsversion')
 
-        pread.assert_called_once_with(['/usr/sbin/rpcinfo', '-p', 'aServer'], quiet=False)
+        pread.assert_called_once_with(['/usr/sbin/rpcinfo', '-s', 'aServer'], quiet=False)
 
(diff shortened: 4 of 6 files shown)


benjamreis:
`rpcinfo -p` has been changed to `rpcinfo -s` because the `-p` option doesn't support ipv6.
https://docs.oracle.com/cd/E88353_01/html/E72487/rpcinfo-8.html

MarkSymsCtx:
In order to be able accept changes to the SM code base we now require all new code to have at least 80% coverage from unit tests. Please ensure that this metric is achieved.

benjamreis:
Sure, so if I understood you correctly, I need to write new tests that covers the IPv6 cases?
Am I correct?

MarkSymsCtx:
> Sure, so if I understood you correctly, I need to write new tests that covers the IPv6 cases?
> Am I correct?

Yes, mostly the new code in util.py as that's the biggest chunk. Ideally all of the code you've modified should be covered by the tests.

benjamreis:
I can easily write a test for new code in `util.py` but for the rest, it'll require an IPv6 with a working NFS target so I don't know how that's possible.

The existing tests doesn't use real wolrd IP, only a dummy host name `aServer`.

benjamreis:
Okay so my utility method is tested, where can i found the code coverage info about my PR?

MarkSymsCtx:
The build will show them but it's not very detailed - https://github.com/xapi-project/sm/pull/547/checks?check_run_id=1880396594
You can run the tests locally with `tests/run_python_unit_tests.sh` and that will create an htmlcov directory with a browsable html coverage report.

benjamreis:
Okay Thx, so i think its good now. :)

MarkSymsCtx:
Sorry, that's not going to be sufficient, it only hits 70% coverage. 
- 4 lines in nfs.py
- 3 lines in NFSSR.py
All the missing lines are in methods which have at least some coverage so there should be some existing unit tests that can be extended or replicated for IPV6 as well.

MarkSymsCtx (changes requested):
More coverage by unit tests will be required to pass build gating.

drivers/NFSSR.py
```diff
@@ -94,8 +94,11 @@ def load(self, sr_uuid):
 
         # Handle optional dconf attributes
         self.transport = DEFAULT_TRANSPORT
+        useIPv6 = util.get_ip_address_family(self.remoteserver) == 6
```
MarkSymsCtx, line 97: This isn't a pythonic variable name. should be `use_ipv6`. Acceptable patterns being  `^[_a-z][a-z0-9_]*$`

MarkSymsCtx: Also adding these additional conditionals has pushed the cognitive complexity of the method over the allowed limits. Probably pushing this bit and the bit below into a `self.set_transport(...)` method would make it more readable.

MarkSymsCtx (changes requested):

benjamreis:
I've made the changes and added some tests, hopefully this will increase the coverage enough.

I used camelCase in my code because some variables where already in camelCase in the existing.

Would it be possible to show in the PR the code coverage of the added code and the cognitive complexity of the code?
It'd made contribution way easier for outsiders since it's really hard to find this info and the rules to follow aren't easily found either.

tests/test_NFSSR.py
```diff
@@ -89,3 +89,8 @@ def test_attach(self, validate_nfsversion, check_server_tcp, _testhost,
                                            timeout=100,
                                            nfsversion='aNfsversionChanged',
                                            retrans=3)
+
+    @mock.patch('NFSSR.Lock', autospec=True)
+    def test_load_ipv6(self, Lock):
```
MarkSymsCtx, line 94: make this `lock` or `mock_lock`, CamelCase isn't PEP compliant syntax

tests/test_nfs.py
```diff
@@ -89,6 +91,17 @@ def test_soft_mount(self, pread, check_server_service, makedirs):
...
 
+    @mock.patch('util.makedirs', autospec=True)
+    @mock.patch('nfs.check_server_service', autospec=True)
+    @mock.patch('util.pread', autospec=True)
+    def test_soft_mount_ipv6(self, pread, check_server_service, makedirs):
+        nfs.soft_mount('mountpoint', 'remoteserver', 'remotepath', 'tcp6',
+                       timeout=None)
+
+        check_server_service.assert_called_once_with('remoteserver')
+        pread.assert_called_once_with(self.get_soft_mount_pread('mount.nfs',
```
MarkSymsCtx, line 102: The scanner doesn't like the duplication of "mount.nfs" but it shouldn't fail the build so happy to leave this as it is without extracting as a constant.

drivers/nfs.py
```diff
@@ -261,12 +263,13 @@ def get_supported_nfs_versions(server):
...
     cv = set()
     try:
-        ns = util.pread2([RPCINFO_BIN, "-p", "%s" % server])
+        ns = util.pread2([RPCINFO_BIN, "-s", "%s" % server])
         ns = ns.split("\n")
         for i in range(len(ns)):
             if ns[i].find("nfs") > 0:
-                cvi = ns[i].split()[1]
-                cv.add(cvi)
+                cvi = ns[i].split()[1].split(",")
```
MarkSymsCtx, line 270: It looks like this section gets no coverage in the tests, it probably didn't before either but given the newer, more complicated processing here it would be good to prove the code matches the intention by having some tests here. It's probably unlikely to get anything other than what is expected but if there are any error responses that need to be considered they should be tested as well.

MarkSymsCtx (changes requested):

tests/test_NFSSR.py
benjamreis: `Lock` is used in all other tests with capital L, won't fail if I change it? I'll try.

MarkSymsCtx: I'm only saying change the parameter to `test_load_ipv6` that's just the mock object that is created by the `@mock.patch` call so that you could if you needed setup up some pre-conditions or responses and check the thing got called as you expect. It doesn't change what the code under test is called.
i.e.
```
@mock.patch('NFSSR.Lock', autospec=True)
    def test_load_ipv6(self, mock_lock):
```

drivers/nfs.py
benjamreis: what happens here is that rpc info returns:
```
# rpcinfo -p <IP>
   program vers proto   port  service
    100000    4   tcp    111  portmapper
    100000    3   tcp    111  portmapper
    100000    2   tcp    111  portmapper
    100000    4   udp    111  portmapper
    100000    3   udp    111  portmapper
    100000    2   udp    111  portmapper
    100005    1   udp  56622  mountd
    100005    1   tcp  42377  mountd
    100005    2   udp  36412  mountd
    100005    2   tcp  60855  mountd
    100005    3   udp  37001  mountd
    100005    3   tcp  53917  mountd
    100003    3   tcp   2049  nfs
    100003    4   tcp   2049  nfs
    100227    3   tcp   2049  nfs_acl
    100003    3   udp   2049  nfs
    100227    3   udp   2049  nfs_acl
    100021    1   udp  35853  nlockmgr
    100021    3   udp  35853  nlockmgr
    100021    4   udp  35853  nlockmgr
    100021    1   tcp  39059  nlockmgr
    100021    3   tcp  39059  nlockmgr
    100021    4   tcp  39059  nlockmgr

# rpcinfo -s <IP>
   program version(s) netid(s)                         service     owner
    100000  2,3,4     local,udp,tcp,udp6,tcp6          portmapper  superuser
    100005  3,2,1     tcp6,udp6,tcp,udp                mountd      superuser
    100003  4,3       udp6,tcp6,udp,tcp                nfs         superuser
    100227  3         udp6,tcp6,udp,tcp                nfs_acl     superuser
    100021  4,3,1     tcp6,udp6,tcp,udp                nlockmgr    superuser
```

But how can i test it without a real target?

MarkSymsCtx: Use the mocks in the unit tests. As an example - https://github.com/xapi-project/sm/blob/master/tests/test_nfs.py#L40 which provides the data for the return of the command

MarkSymsCtx (approved):

MarkSymsCtx:
> I've made the changes and added some tests, hopefully this will increase the coverage enough.

Yes, thanks, all good now

> I used camelCase in my code because some variables where already in camelCase in the existing.

There's lots of things wrong with the existing code, we're slowly working through things improving them and trying to make sure that new code is "done right"

> Would it be possible to show in the PR the code coverage of the added code and the cognitive complexity of the code?
> It'd made contribution way easier for outsiders since it's really hard to find this info and the rules to follow aren't easily found either.
Probably not, we have a SonarQube server in the main Citrix build process and it is this that is doing build to build coverage comparisons and scanning for the complexity. Flake8 will do some of the PEP compliance bits but it's not as comprehensive as SonarQube.

TimSmithCtx (approved):

BenSimsCitrix (approved):
LGTM

Issue #511, opened 2020-08-27
nfs sr-probe fails with QNAP NFS devices


olivierlambert:
Hi!

We had numerous reports of users having issues to create an NFS SR when using QNAP devices.

So it seems the problem is because the probe request (even on `xe`) doesn't return anything, eg:

```
xe sr-probe type=nfs device-config:server=192.168.0.15
Error code: SR_BACKEND_FAILURE_101
Error parameters: , The request is missing the serverpath parameter,
```

It should be:
```
xe sr-probe type=nfs device-config:server=192.168.0.15
Error code: SR_BACKEND_FAILURE_101
Error parameters: , The request is missing the serverpath parameter,
<nfs-exports>
	<Export>
		<Target>192.168.0.15</Target>
		<Path>/foo/bar</Path>
		<Accesslist>(everyone)</Accesslist>
	</Export>
</nfs-exports>
```

After some investigation, the common point was using a QNAP device, and having `showmount` to return share lists without any "permissions". Eg, an expected `showmount` would be:

```
# showmount -e 10.0.1.197
Export list for 10.0.1.197:
/mnt/tank/backups/Xen 10.0.1.0
/mnt/tank/ISO         10.0.1.1
/mnt/tank/home/kevdog 10.0.1.1
```

But with QNAP devices, for example:
```
# showmount -e 172.18.9.34
Export list for 172.18.9.34:
/vm      
/Public  
/Web     
```

I suppose this is the reason of the issue: in NFS probe function, it expects to have another row with the permission (IP or "everyone"). QNAP told us there's no possibility to modify permission for the NFS, so we can't test on that side.

Problem might be around here: https://github.com/xapi-project/sm/blob/46a8c7a597a3671c4120db68fcc7efe1c2bee35c/drivers/nfs.py#L187

It might be doable to manage this case when there's no ACLs visible for a share. At least, that's my theory, your input is welcome :+1:

edwintorok:
Which QNAP device model? IIRC we do have some QNAP devices internally that do work, and there is a list of tested QNAP devices here: http://hcl.xensource.com/storage/?vendor=65
Or is this due to a non-default configuration?

olivierlambert:
Reports are coming (cf issues pointing to here) from QNAP TS-873U-RP (4.3.5.0760), and various QNAP OS version (QTS 4.4.1, and upper).

[...]

PR #548, opened 2021-02-12
If no NFS ACLs provided, assume everyone

benjamreis:
Signed-off-by: BenjiReis <benjamin.reis@vates.fr>

Solves: https://github.com/xapi-project/sm/issues/511

As detailed in the issue, some devices (mostly QNAS devices) do not provide with NFS ACLs in `showmount -e` return, in this case we need to assume everyone is accepted, therefore `access = "*"`.

### DIFF
--- a/drivers/nfs.py
+++ b/drivers/nfs.py
@@ -205,7 +205,11 @@ def scan_exports(target):
         textnode = dom.createTextNode(target)
         subentry.appendChild(textnode)
 
-        (path, access) = val.split()
+        # Access is not always provided by showmount return
+        # If none is provided we need to assume "*"
+        array = val.split()
+        path = array[0]
+        access = array[1] if len(array) >= 2 else "*"
         subentry = dom.createElement("Path")
         entry.appendChild(subentry)
         textnode = dom.createTextNode(path)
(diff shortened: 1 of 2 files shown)


drivers/nfs.py
```diff
@@ -203,7 +203,11 @@ def scan_exports(target):
         textnode = dom.createTextNode(target)
         subentry.appendChild(textnode)
 
-        (path, access) = val.split()
+        # Access is not always provided by showmount return
+        # It none is provided we need to assume "*"
```
benjamreis, line 207:
```suggestion
        # If none is provided we need to assume "*"
```
```suggestion
        # It none is provided we need to assume "*"
```

benjamreis:
Can `showmount` be mocked to fake a result with no `access` in it?

MarkSymsCtx:
> Can `showmount` be mocked to fake a result with no `access` in it?

Of course, if you know what the response looks like, just make the return value from the mock look like that.

benjamreis:
Okay thanks, I'll add a test then :)

benjamreis:
Test added and successfull, this is good to review, and that's one more nfs method covered by the test! :+1:

benjamreis:
Hi!

Any changes required? Or something is missing? Tests and test coverage should be fine.

Regards!

TimSmithCtx (approved):

PR #552, opened 2021-04-29
Fix regression added by XSI-915:

Wescoeur:
If device is missing during LVM/EXT SR creation we have this error:
```
Error code: SR_BACKEND_FAILURE_1200
Error parameters: , 'LVHDSR' object has no attribute 'root',
```

With this fix we retrieve a more explicit message:
```
Error code: SR_BACKEND_FAILURE_90
Error parameters: , The request is missing the device parameter,
```

### DIFF
--- a/drivers/EXTSR.py
+++ b/drivers/EXTSR.py
@@ -64,6 +64,10 @@ class EXTSR(FileSR.FileSR):
         self.ops_exclusive = FileSR.OPS_EXCLUSIVE
         self.lock = Lock(vhdutil.LOCK_TYPE_SR, self.uuid)
         self.sr_vditype = SR.DEFAULT_TAP
+        if type(self) == EXTSR and (
+            'device' not in self.dconf or not self.dconf['device']
+        ):
+            raise xs_errors.XenError('ConfigDeviceMissing')
 
         self.path = os.path.join(SR.MOUNT_BASE, sr_uuid)
         self.vgname = EXT_PREFIX + sr_uuid
--- a/drivers/LVHDSR.py
+++ b/drivers/LVHDSR.py
@@ -147,6 +147,10 @@ class LVHDSR(SR.SR):
 
     def load(self, sr_uuid):
         self.ops_exclusive = OPS_EXCLUSIVE
+        if type(self) == LVHDSR and (
+            'device' not in self.dconf or not self.dconf['device']
+        ):
+            raise xs_errors.XenError('ConfigDeviceMissing')
 
         self.isMaster = False
         if 'SRmaster' in self.dconf and self.dconf['SRmaster'] == 'true':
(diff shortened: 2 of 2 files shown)


d003kk (commented):
Thanks for finding this.
I'm running this over our test suite to ensure no further issues.

We need some unit tests to enforce this.

TimSmithCtx (approved):

BenSimsCitrix:
This change causes further regressions. For example 

 xe sr-create type=lvmoiscsi content-type= physical-size=0 name-label=test-iscsi shared=true device-config:target=10.9.9.9 device-config:targetIQN=iqn.test:LUNid=0 device-config:SCSIid=123 username=root password=***

We need further thought here.

Wescoeur:
@BenSimsCitrix I'm not sure to understand, what's the ouput of your command? My change or the XSI-915 commit causes other regressions?

Because in theory this fix doesn't impact the `lvmoiscsi` driver, there is a specific `load` method in this last one, so the `load` parent in `LVHDSR` should not be called because it is overriden.

MarkSymsCtx:
> @BenSimsCitrix I'm not sure to understand, what's the ouput of your command? My change or the XSI-915 commit causes other regressions?
> 
> Because in theory this fix doesn't impact the `lvmoiscsi` driver, there is a specific `load` method in this last one, so the `load` parent in `LVHDSR` should not be called because it is overriden.

Your change introduces regressions in core functionality, notably that the above command does not complete successfully.

Wescoeur:
Indeed I had not seen the load method was explicitly called, sorry... I force pushed another solution, but I don't know if this problem can be solved in a more elegant way. In any case, testing the type using the builtin `type` function fixes the problem.

MarkSymsCtx:
@Wescoeur I have to say that I'm none too keen on this way of resolving the issue. Unfortunately our team is currently swamped with some other issues that we're working on so it will take us a bit of time to get back to looking at this to see if we can get a better solution.

MarkSymsCtx:
@Wescoeur do you have some tests (even if it's just a manual script) that reproduced this issue? I assume you had some xcp-ng bugs filed for this. If you give us a quick, if you do this and this then bad things happen it would really help in nailing this down.

Wescoeur:
I just pushed a test in our repo to ensure the bug is not present: https://github.com/xcp-ng/xcp-ng-tests/pull/34
But we must add several tests regarding other lvm based drivers.

MarkSymsCtx:
> I just pushed a test in our repo to ensure the bug is not present: [xcp-ng/xcp-ng-tests#34](https://github.com/xcp-ng/xcp-ng-tests/pull/34)
> But we must add several tests regarding other lvm based drivers.

Thanks, a unit test for this is great. But just so I'm clear, the original issue showed up when doing something like
```
xe sr-create name-label="XXXXX" type=ext
```
and not providing the `device-config:device=XXXX` bit?

Wescoeur:
Yes, exactly, the problem is here if the device is not given in the command line.

MarkSymsCtx:
This issue should be resolved by https://github.com/xapi-project/sm/pull/555

PR #556, opened 2021-06-09
CA-355401: make post attach scan best effort and report errors

MarkSymsCtx:


TimSmithCtx (dismissed):

drivers/SR.py
```diff
@@ -301,7 +301,15 @@ def after_master_attach(self, uuid):
         Return:
           None
         """
-        self.scan(uuid)
+        try:
+            self.scan(uuid)
+        except Exception as e:
+            util.SMlog("Error in SR.after_master_attach %s" % e)
+            msg_name = "Post attach scan failed"
```
edwintorok, line 308: I think the message names are usually uppercase with underscores, and then XenCenter can have a localized string as mapping for it. Not sure whether spaces would be allowed

MarkSymsCtx: OK, I copied an existing one, which is probably wrong.

TimSmithCtx (approved):

rosslagerwall (approved):

PR #558, opened 2021-06-15
Remove dead code from tapdisk-pause

MarkSymsCtx:
This has to be dead as it references the non-existant self variable
in a global context. There is an almost identical copy of the
method inside the Tapdisk class as an instance method which can use
self.

### DIFF
--- a/drivers/tapdisk-pause
+++ b/drivers/tapdisk-pause
@@ -73,24 +73,6 @@ def _mkphylink(sr_uuid, vdi_uuid, path):
     util.pread2(cmd)
     return path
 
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
This is the actually used, correct version - https://github.com/xapi-project/sm/pull/558/files#diff-c5138be9cdacb74cdcfd4fe358e84be2f25f66933aa1403e834e2fbf6d7515cdR113

rosslagerwall (approved):

edwintorok (approved):

TimSmithCtx (approved):

PR #559, opened 2021-06-30
CA-356234: don't run repair if parent is raw

MarkSymsCtx:

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -885,10 +885,11 @@ class VDI:
             parent = ""
             try:
                 parent = vhdutil.getParent(self.path, lambda x: x.strip())
-                # Repair error is logged and ignored. Error reraised later
-                util.SMlog('Coalesce failed on %s, attempting repair on ' \
-                           'parent %s' % (self.uuid, parent))
-                vhdutil.repair(parent)
+                if not parent.raw:
+                    # Repair error is logged and ignored. Error reraised later
+                    util.SMlog('Coalesce failed on %s, attempting repair on ' \
+                               'parent %s' % (self.uuid, parent))
+                    vhdutil.repair(parent)
             except Exception as e:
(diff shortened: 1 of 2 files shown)


TimSmithCtx (approved):

drivers/cleanup.py
```diff
@@ -885,10 +885,11 @@ def _coalesceVHD(self, timeOut):
             parent = ""
             try:
                 parent = vhdutil.getParent(self.path, lambda x: x.strip())
-                # Repair error is logged and ignored. Error reraised later
-                util.SMlog('Coalesce failed on %s, attempting repair on ' \
-                           'parent %s' % (self.uuid, parent))
-                vhdutil.repair(parent)
+                if not parent.raw:
```
edwintorok, line 888: I didn't know you could have VHD chains built with raw elements, I thought that being raw meant you can take no snapshots and have no trees. Although I think you can have exactly one raw element in a VHD chain, its base and then other VHD children use their bitmaps to say which blocks they've modified.
Since for LVM we already take up the full space anyway (no thin provisioning) can we always make the first non-empty element in the VHD chain raw as a performance optimization?

edwintorok (approved):

drivers/cleanup.py
MarkSymsCtx: > I didn't know you could have VHD chains built with raw elements, I thought that being raw meant you can take no snapshots and have no trees. Although I think you can have exactly one raw element in a VHD chain, its base and then other VHD children use their bitmaps to say which blocks they've modified.
> Since for LVM we already take up the full space anyway (no thin provisioning) can we always make the first non-empty element in the VHD chain raw as a performance optimization?

This is strictly a legacy mode. It means that the ultimate top level parent must always be fully provisioned whereas the non-legacy mode full VHD tree will truncate the parent to the smallest volume possible when the first snapshot is taken.

PR #561, opened 2021-07-05
CA-356102: retry LV commands when 'Incorrect checksum in metadata' re…

MarkSymsCtx:
…ported

### DIFF
--- a/drivers/lvutil.py
+++ b/drivers/lvutil.py
@@ -138,6 +138,34 @@ class LvmLockContext(object):
             self.lock.release()
 
 
+LVM_RETRY_ERRORS = [
+    "Incorrect checksum in metadata area header"
+]
+
+
+def lvmretry(func):
+    def check_exception(exception):
+        retry = False
+        for error in LVM_RETRY_ERRORS:
+            if error in exception.reason:
+                retry = True
+        return retry
+
+    def decorated(*args, **kwargs):
+        for i in range(LVM_FAIL_RETRIES):
+            try:
+                return func(*args, **kwargs)
+            except util.CommandException as ce:
+                retry = check_exception(ce)
+                if not retry or (i == LVM_FAIL_RETRIES - 1):
+                    raise
+
+                time.sleep(1)
+
+    decorated.__name__ = func.__name__
+    return decorated
+
+
 def cmd_lvm(cmd, pread_func=util.pread2, *args):
     """ Construct and run the appropriate lvm command.
 
@@ -550,6 +578,7 @@ def setActiveVG(path, active):
     text = cmd_lvm([CMD_VGCHANGE, "-a" + val, path])
 
 
+@lvmretry
 def create(name, size, vgname, tag=None, size_in_percentage=None):
     if size_in_percentage:
         cmd = [CMD_LVCREATE, "-n", name, "-l", size_in_percentage, vgname]
@@ -576,6 +605,7 @@ def remove(path, config_param=None):
     _lvmBugCleanup(path)
 
 
+@lvmretry
 def _remove(path, config_param=None):
     CONFIG_TAG = "--config"
     cmd = [CMD_LVREMOVE, "-f", path]
@@ -584,10 +614,12 @@ def _remove(path, config_param=None):
     ret = cmd_lvm(cmd)
 
 
+@lvmretry
 def rename(path, newName):
     cmd_lvm([CMD_LVRENAME, path, newName], pread_func=util.pread)
 
 
+@lvmretry
 def setReadonly(path, readonly):
     val = "r"
     if not readonly:
@@ -600,6 +632,7 @@ def exists(path):
     return rc == 0
 
 
+@lvmretry
 def setSize(path, size, confirm):
     sizeMB = size / (1024 * 1024)
     if confirm:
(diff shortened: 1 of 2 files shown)


MarkSymsCtx:
Errors seen in logs of the form
```
Jun 26 08:07:23 <host> SM: [19375] ['/sbin/lvchange', '-ay', '/dev/VG_XenStorage-94d5c7de-3bee-e7c2-8aeb-d609e7dcd358/VHD-b1612f97-8a16-4cdd-9d7d-a1bf3283acc9']
Jun 26 08:07:23 <host> SM: [19375] FAILED in util.pread: (rc 5) stdout: '', stderr: '  Incorrect checksum in metadata area header on /dev/sdb at 4096
Jun 26 08:07:23 <host> SM: [19375]   Failed to read mda header from /dev/sdb
Jun 26 08:07:23 <host> SM: [19375]   Failed to scan VG from /dev/sdb
Jun 26 08:07:23 <host> SM: [19375]   Volume group "VG_XenStorage-94d5c7de-3bee-e7c2-8aeb-d609e7dcd358" not found
Jun 26 08:07:23 <host> SM: [19375]   Cannot process volume group VG_XenStorage-94d5c7de-3bee-e7c2-8aeb-d609e7dcd358
Jun 26 08:07:23 <host> SM: [19375] '
```
Subsequent calls to `lvchange -ay` on the same VG on this host succeeded which suggests this is a host to host race on updating the VG metadata as the volume state is changed.

olivierlambert:
That's really interesting, thanks for the detailed explanation! (I think we've seen it from time to time without being able to grasp the root cause)

MarkSymsCtx:
> That's really interesting, thanks for the detailed explanation! (I think we've seen it from time to time without being able to grasp the root cause)

I think it probably quite rare, hence trying to be really heavy on the unit tests as whilst we did see this in our automated tests it's the first time I have seen it and given the amount of testing that's running 24/7 that's quite surprising. I certainly don't have a good means to efficiently reproduce it in real-world conditions to categorically prove that the fix is good and will be sufficient.

olivierlambert:
In general, this is the kind of experience/errors we had with slow iSCSI SAN (eg too many VDIs on HDD SAN) or slow/flaky network. Hard to really reproduce on a "working" QA environment. We probably have far more "smaller" customers, meaning in general potential bad config or bad/cheap hardware. That might explain why we experienced that in the past.

Those problem are the worst to debug (race condition in specific environments) and there's little we can do.

drivers/lvutil.py
```diff
@@ -138,6 +138,34 @@ def __exit__(self, exc_type, value, traceback):
             self.lock.release()
 
 
+LVM_RETRY_ERRORS = [
+    "Incorrect checksum in metadata area header"
```
edwintorok, line 142: Isn't this an indication that we're missing a lock somewhere, or not flushing updates properly?
Retrying might fix the issue on read-only operations, but don't we also have an issue where 2 writers could race and end up in a lost update if the 2nd writer started out with slightly older information and overwrote the changes made by the other writer?

drivers/lvutil.py
```diff
@@ -138,6 +138,34 @@ def __exit__(self, exc_type, value, traceback):
...
+    def check_exception(exception):
+        retry = False
+        for error in LVM_RETRY_ERRORS:
+            if error in exception.reason:
+                retry = True
+        return retry
+
+    def decorated(*args, **kwargs):
+        retry = True
+        while retry:
```
edwintorok, line 156: This will result in an infinite loop if the metadata is truly corrupted. Probably needs a counter and give up after N retries (otherwise we'll just fill /var/log whenever we got corruption, potentially even loosing the original reason due to aggressively triggering log rotation).

drivers/lvutil.py
```diff
@@ -138,6 +138,34 @@ def __exit__(self, exc_type, value, traceback):
...
+            retry = False
+            try:
+                func(*args, **kwargs)
```
edwintorok, line 159: Once a counter is added I think we should log a message here if we succeed with counter > 1. Then we can have a healthcheck in XenRT to detect this condition and track how likely it is.

edwintorok (changes requested):

drivers/lvutil.py
edwintorok: (marked as 'changes required' to fix the infinite loop)

drivers/lvutil.py
MarkSymsCtx: Absolutely we're missing a lock, we're missing a lock because we abuse LVM and use it from multiple hosts without a corosync cluster. This is an immense pile of technical debt that we can only patch up and bandage and hope and pray that thing hang together.

drivers/lvutil.py
```diff
@@ -550,6 +578,7 @@ def setActiveVG(path, active):
     text = cmd_lvm([CMD_VGCHANGE, "-a" + val, path])
 
 
+@lvmretry
```
rosslagerwall, line 581: Would it be better to decorate `cmd_lvm` instead of all its callers?

rosslagerwall (approved):

drivers/lvutil.py
MarkSymsCtx: > Would it be better to decorate `cmd_lvm` instead of all its callers?

If it was all the callers, yes, but it's not, so no.

TimSmithCtx (approved):

PR #562, opened 2021-07-08
CA-356411: correct check for session

MarkSymsCtx:


rosslagerwall (dismissed):

edwintorok (approved):
I hope this won't reintroduce the problems we had when plugging 100s of SRs on boot. I recommend running the 'manysrs' testcase after this is built.

MarkSymsCtx:
> I hope this won't reintroduce the problems we had when plugging 100s of SRs on boot. I recommend running the 'manysrs' testcase after this is built.

Running now.

edwintorok:
Determined that `not s` invokes `__nonzero__` API call in XAPI (the xmlrpc proxy forwards everything, including special method calls), whereas `s is None` doesn't invoke any builtin methods.

stormi:
> Determined that `not s` invokes `__nonzero__` API call in XAPI (the xmlrpc proxy forwards everything, including special method calls), whereas `s is None` doesn't invoke any builtin methods.

Maybe this subtlety would deserve a comment in the code, or at least a mention in the commit message?

stormi:
(By the way as an external observer it's really hard to understand the reasons for any changes when all we have in the commit messages is a Citrix internal CA number)

MarkSymsCtx:
> > Determined that `not s` invokes `__nonzero__` API call in XAPI (the xmlrpc proxy forwards everything, including special method calls), whereas `s is None` doesn't invoke any builtin methods.
> 
> Maybe this subtlety would deserve a comment in the code, or at least a mention in the commit message?

It should but it actually belongs on the Xapi Python API docs.

rosslagerwall (approved):

TimSmithCtx (approved):

edwintorok (approved):

PR #563, opened 2021-07-13
CA-356645: use "self.session is None" not "self.session == None"

MarkSymsCtx:
Python runtime converts use of == to a call to "__eq__" on the object
and this is not implemented by the Xapi session XMLRPC proxy. After
https://github.com/xapi-project/xen-api/pull/4464 this code causes an
exception.

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -281,7 +281,7 @@ class XAPI:
     def __init__(self, session, srUuid):
         self.sessionPrivate = False
         self.session = session
-        if self.session == None:
+        if self.session is None:
             self.session = self.getSession()
             self.sessionPrivate = True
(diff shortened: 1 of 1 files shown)


edwintorok (dismissed):

TimSmithCtx (dismissed):

edwintorok:
Few more instances to fix:
`drivers/SMBSR.py:                self.session == ""):`

MarkSymsCtx:
> One more instance to fix:
> `drivers/SMBSR.py: self.session == ""):`
> 
> Looking through `git grep 'self.session '` thats the only problematic one I can spot

No as the full code for that one is `(isinstance(self.session, str) and self.session == "")`, there's a line break in the middle of the line though.

TimSmithCtx (approved):

edwintorok (approved):

PR #565, opened 2021-07-21
CA-356761: ensure that children are always reloaded on relink

MarkSymsCtx:
Extend allowed wait in garbage collector for VDI activating.

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -940,7 +940,7 @@ class VDI:
         if len(self.children) == 0:
             retries = 0
             try:
-                while retries < 10:
+                while retries < 15:
                     retries += 1
                     if self.getConfig(VDI.DB_VDI_ACTIVATING) is not None:
                         Util.log("VDI %s is activating, wait to relink" %
@@ -954,7 +954,7 @@ class VDI:
                                      self.uuid)
                         else:
                             return
-                    time.sleep(1)
+                    time.sleep(2)
 
                 raise util.SMException("Failed to tag vdi %s for relink" % self)
             except XenAPI.Failure as e:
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
We saw failures in testing where the backoff in `_tagChildrenForRelink` expired because a VDI was activating and that took longer than the grace period. This then left the VDI marked with `relinking` which prevented future activations. Ensure we never leave without clearing the tags, by calling `_reloadChildren`. Also increase the grace delay to reduce the chances of this happening anyway.

drivers/cleanup.py
```diff
@@ -940,7 +940,7 @@ def _tagChildrenForRelink(self):
         if len(self.children) == 0:
             retries = 0
             try:
-                while retries < 10:
+                while retries < 15:
```
edwintorok, line 943: The use of DB_VDI_ACTIVATING below is confusing, because it is only ever read when doing a `git grep`. blktap2.py has an `activating` key, but doesn't use this constant, can the constants be shared between the 2 files? Same for relinking

rosslagerwall (approved):

drivers/cleanup.py
edwintorok: The code below uses 2 separate keys to implement a lock, I can see why it is done this way (due to how setConfig is implemented), but it might be better to use a single 'lock' field and instead of remove+add, just use 'add' and catch the exception saying Duplicate that the field is already set (that means someone else is holding a lock). And you'd only ever do a 'removeFromConfigVDI' if you are the one currently holding a lock. That means we wouldn't need the get1 + set2 + get1 + del2 logic at all. For additional safety the value should be globally unique (e.g. host UUID + PID): XAPI only responds with Duplicate if the value is different.

edwintorok (approved):

drivers/cleanup.py
MarkSymsCtx: Refactoring these constants is a fairly sizeable job as they currently exist only in the VDI subclass in the garbage collection code. But yes, that is definitely something that would improve the readability of the code.

The other point about using a single lock field would need to be looked at carefully, how good are the atomicy guarantees for setConfig? (Maybe that doesn't change anything significantly though)

TimSmithCtx (approved):

PR #567, opened 2021-07-28
CA-334762/CA-356983: fix VDI import for storage driver domains

edwintorok:
Details explained in commit messages: udev is launching various commands when xvd* devices are plugged into Dom0 and this prevents unplug with a device in use error. Plug/unplug is needed for VDI import to work.

Tested with a storage driver domain, and also added some cleanups for td/nbd (there were udev rules to ignore these devices, but not everywhere, e.g. I was still getting some multipath commands launched for them).
Still need to test the td/nbd side of changes, hence the draft PR.

### DIFF
--- a/udev/58-xapi.rules
+++ b/udev/58-xapi.rules
@@ -3,6 +3,8 @@
 # Skip block devices - no networks in here
 SUBSYSTEM!="block", GOTO="end_xapi"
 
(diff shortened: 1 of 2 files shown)


edwintorok:
Tests look good so far.

rosslagerwall (approved):

MarkSymsCtx (approved):

TimSmithCtx (approved):

PR #568, opened 2021-09-02
CA-352880: when deleting an HBA SR remove the kernel devices

MarkSymsCtx:
If this is not done, in particular with a Fibre Channel SR where
the LUN mapping may be subsequently removed kernel errors can be
encnountered until the host is rebooted if another HBA SR is
scanned.

e.g.

Mar 23 13:43:10 <host> kernel: [446324.856765] sd 1:0:2:0: [sdd] tag#0 FAILED Result: hostbyte=DID_OK driverbyte=DRIVER_SENSE
Mar 23 13:43:10 <host> kernel: [446324.856775] sd 1:0:2:0: [sdd] tag#0 Sense Key : Illegal Request [current]
Mar 23 13:43:10 <host> kernel: [446324.856778] sd 1:0:2:0: [sdd] tag#0 Add. Sense: Logical unit not supported
Mar 23 13:43:10 <host> kernel: [446324.856784] sd 1:0:2:0: [sdd] tag#0 CDB: Read(10) 28 00 00 00 00 00 00 00 80 00

If the host is rebooted between the SR being deleted and the LUN
mapping being removed then this error will recur until the host
is rebooted again.

### DIFF
--- a/drivers/LVHDoHBASR.py
+++ b/drivers/LVHDoHBASR.py
@@ -195,6 +195,17 @@ class LVHDoHBASR(LVHDSR.LVHDSR):
             except:
                 pass
 
+    def _remove_device_nodes(self):
+        """
+        Remove the kernel device nodes
+        """
+        nodes = glob.glob('/dev/disk/by-scsid/%s/*' % self.SCSIid)
+        util.SMlog('Remove_nodes, nodes are %s' % nodes)
+        for node in nodes:
+            with open('/sys/block/%s/device/delete' %
+                      (os.path.basename(node)), 'w') as f:
+                f.write('1\n')
+
     def delete(self, sr_uuid):
         self._pathrefresh(LVHDoHBASR)
         try:
@@ -202,6 +213,7 @@ class LVHDoHBASR(LVHDSR.LVHDSR):
         finally:
             if self.mpath == "true":
                 self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
+            self._remove_device_nodes()
 
     def vdi(self, uuid):
         return LVHDoHBAVDI(self, uuid)
--- a/drivers/LVHDoHBASR.py
+++ b/drivers/LVHDoHBASR.py
@@ -107,7 +107,7 @@ class LVHDoHBASR(LVHDSR.LVHDSR):
                 raise xs_errors.XenError('ConfigSCSIid')
 
         self.SCSIid = self.dconf['SCSIid']
-        LVHDSR.LVHDSR.load(self, sr_uuid)
+        super(LVHDoHBASR, self).load(sr_uuid)
 
     def create(self, sr_uuid, size):
         self.hbasr.attach(sr_uuid)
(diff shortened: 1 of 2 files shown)


MarkSymsCtx:
This will need some unit tests adding to meet the code coverage requirements.

rosslagerwall:
```
If the host is rebooted between the SR being deleted and the LUN
mapping being removed then this error will recur until the host
is rebooted again.
```

Does the change here help this particular issue? Removing the block device surely only has an effect until the host is rebooted?

drivers/LVHDoHBASR.py
```diff
@@ -195,13 +195,25 @@ def detach(self, sr_uuid):
...
                 pass
 
+    def _remove_device_nodes(self):
+        """
+        Remove the kernel device nodes
+        """
+        nodes = glob.glob('/dev/disk/by-scsid/%s/*' % self.SCSIid)
+        util.SMlog('Remove_nodes, nodes are %s' % nodes)
+        for node in nodes:
+            with open('/sys/block/%s/device/delete' %
```
rosslagerwall, line 205: If the block device is deleted, how would you recreate the SR without a host reboot?

drivers/LVHDoHBASR.py
```diff
@@ -195,13 +195,25 @@ def detach(self, sr_uuid):
...
+                f.write('1\n')
+
     def delete(self, sr_uuid):
         self._pathrefresh(LVHDoHBASR)
         try:
             LVHDSR.LVHDSR.delete(self, sr_uuid)
         finally:
             if self.mpath == "true":
                 self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
+            self._remove_device_nodes()
```
rosslagerwall, line 216: Could this fail and leave the SR in an unexpected state?

drivers/LVHDoHBASR.py
MarkSymsCtx: The SR create rescans the HBA and "rediscovers" the device.

MarkSymsCtx:
> ```
> If the host is rebooted between the SR being deleted and the LUN
> mapping being removed then this error will recur until the host
> is rebooted again.
> ```
> 
> Does the change here help this particular issue? Removing the block device surely only has an effect until the host is rebooted?

Sort of. Yes, it only takes effect for the duration of the current execution cycle, but that means it can cope with
* Destroy SR
* Unmap LUN/retire SAN
* Scan other SR - no errors
a reboot after that won't see the device as it's been upmapped from the SAN. But if you destroy the SR, then reboot the host and then unmap the LUN then you'll still get kernel errors until you do another host reboot.

drivers/LVHDoHBASR.py
MarkSymsCtx: No, it either succeeds or leaves things as they are today.

rosslagerwall:
> a reboot after that won't see the device as it's been upmapped from the SAN. But if you destroy the SR, then reboot the host and then unmap the LUN then you'll still get kernel errors until you do another host reboot.

I see, that makes sense.

rosslagerwall (approved):

TimSmithCtx (approved):

PR #569, opened 2021-09-02
CA-327302: report signals in CommandException

MarkSymsCtx:
Negative process exit codes indicate that the process terminated
with a signal so report these separately.

### DIFF
--- a/drivers/util.py
+++ b/drivers/util.py
@@ -78,11 +78,18 @@ class SMException(Exception):
 
 
 class CommandException(SMException):
+    def error_message(self, code):
+        if code > 0:
+            return os.strerror(code)
+        elif code < 0:
+            return "Signalled %s" % (abs(code))
+        return "Success"
+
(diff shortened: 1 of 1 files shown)


drivers/util.py
```diff
@@ -78,11 +78,18 @@ class SMException(Exception):
 
 
 class CommandException(SMException):
+    def error_message(self, code):
+        if code > 0:
+            return os.strerror(code)
+        elif code < 0:
+            return "Signalled %s" % (abs(code))
```
rosslagerwall, line 85: It would be helpful to use `signal.strsignal()` if we could use Python 3.8.

rosslagerwall (approved):

drivers/util.py
MarkSymsCtx: that it would, but a lot more work required to get this to be able to run in Python 3.8

TimSmithCtx (approved):

PR #570, opened 2021-09-27
CA-359304: ignore relinking and activating tags in update_sm_config

MarkSymsCtx:
These keys are used ephemerally as part of the GC chain relink and
blktap activation paths to interlock the two processes and may be removed
while the SR scan is occuring.

### DIFF
--- a/drivers/VDI.py
+++ b/drivers/VDI.py
@@ -507,10 +507,17 @@ class VDI(object):
 
     def _db_update_sm_config(self, ref, sm_config):
         import cleanup
+        # List of sm-config keys that should not be modifed by db_update
+        smconfig_protected_keys = [
+            cleanup.VDI.DB_VDI_PAUSED,
+            cleanup.VDI.DB_VHD_BLOCKS,
+            cleanup.VDI.DB_VDI_RELINKING,
+            cleanup.VDI.DB_VDI_ACTIVATING]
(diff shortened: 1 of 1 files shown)


drivers/VDI.py
```diff
@@ -519,9 +522,12 @@ def _db_update_sm_config(self, ref, sm_config):
                 self.sr.session.xenapi.VDI.add_to_sm_config(ref, key, val)
 
         for key in current_sm_config.keys():
-            if key.startswith("host_") or \
-                key in ["paused", cleanup.VDI.DB_VHD_BLOCKS] or \
-                key in self.sm_config_keep:
+            if (key.startswith("host_") or
```
edwintorok, line 525: we used to have backslashes here, and now we don't anymore, I assume that is fine?

edwintorok (approved):

drivers/VDI.py
MarkSymsCtx: I added () which is a nicer way of keeping the boolean contained.

MarkSymsCtx: But, I did forget that I said I would move the list of keys to a single constant, so will update with that

rosslagerwall (approved):

TimSmithCtx (approved):

PR #572, opened 2021-10-04
CA-359621: fix regression in error handler

MarkSymsCtx:
Parent returned from vhdutil.getParent is the path of the parent not a
VDI object. Lookup the VDI before dereferencing the raw attribute. Add
fault injection fist point to allow for testing.

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -537,6 +537,10 @@ class VDI:
         self._vdiRef = None
         self._clearRef()
 
+    @staticmethod
+    def extractUuid(path):
+        raise NotImplementedError("Implement in sub class")
+
     def load(self):
         """Load VDI info"""
         pass  # abstract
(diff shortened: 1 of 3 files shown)


TimSmithCtx (dismissed):

drivers/cleanup.py
```diff
@@ -537,6 +537,10 @@ def __init__(self, sr, uuid, raw):
         self._vdiRef = None
         self._clearRef()
 
+    def extractUuid(path):
+        raise NotImplementedError("Implement in sub class")
+    extractUuid = staticmethod(extractUuid)
```
rosslagerwall, line 542: Isn't it simpler to use the @staticmethod decorator and avoid using of `.__class__` below?

rosslagerwall (approved):

drivers/cleanup.py
MarkSymsCtx: I don't think that the decorator avoid the need for `__class__` and this was keeping the code consistent with the existing instance in FileVDI. But, I can try it.

MarkSymsCtx: Ok, I stand corrected. That does seem to work correctly.

TimSmithCtx (approved):

rosslagerwall (approved):

PR #574, opened 2021-10-06
Add fallback support for network filers which do not support hardlink operations

MarkSymsCtx:
After a customer report of snapshot failures subsequent to https://github.com/xapi-project/sm/commit/f28cd078e7e166fbfba8957e49bea7422004443c which were found to be due to the filer in use not supporting the expected hardlink operation.

Detect this condition at SR.attach and mark the sm_config for SR if the hardlink operation fails and then use the previous mechanism of renames, albeit that this cannot be guaranteed to be crash consistency.

### DIFF
--- a/drivers/FileSR.py
+++ b/drivers/FileSR.py
@@ -784,6 +784,13 @@ class FileVDI(VDI.VDI):
         util.SMlog("FileVDI._unlink %s" % (path))
         os.unlink(path)
 
+    def _create_new_parent(self, src, newsrc):
+        sr_sm_config = self.session.xenapi.SR.get_sm_config(self.sr.sr_ref)
+        if SharedFileSR.NO_HARDLINK_SUPPORT in sr_sm_config:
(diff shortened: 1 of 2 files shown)


drivers/FileSR.py
```diff
@@ -1082,6 +1102,49 @@ def _delete_cbt_log(self):
...
+    """
+    NO_HARDLINK_SUPPORT = "no_hardlinks"
+
+    def _raise_hardlink_error(self):
+        raise OSError(542, "Unknown error 524")
+
+    def _check_hardlinks(self):
+        test_name = os.path.join(self.path, str(uuid4()))
+        f = open(test_name, 'ab')
+        f.close
```
rosslagerwall, line 1118: `f.close()`

Or shorter:

`open(test_name, 'ab').close()`

MarkSymsCtx: Re-running the test.

rosslagerwall (approved):

PR #578, opened 2021-11-26
CP-38670: use monotonic time instead of time.time()

MarkSymsCtx:
This reduces the chances of failures due to the system clock changing,
for instance by NTP/chronyd moving the system time to sync to the server.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -5,3 +5,4 @@ coverage
 astroid==1.4.9
 pylint==1.5
 bitarray
+monotonic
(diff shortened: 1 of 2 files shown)


rosslagerwall (approved):

edwintorok (approved):
There are more instances of `time.time()` used as a delta (see lvutil and LVHDSR), but AFAICT only used to log a warning or not and wouldn't have a functional difference, you could change those 2 to use monotonic if you want.
And the one in trimutil is used to record as a timestamp the last time it ran, so that one should probably be left as is.

MarkSymsCtx:
> There are more instances of `time.time()` used as a delta (see lvutil and LVHDSR), but AFAICT only used to log a warning or not and wouldn't have a functional difference, you could change those 2 to use monotonic if you want. And the one in trimutil is used to record as a timestamp the last time it ran, so that one should probably be left as is.

I think we're less bothered by those and potentially that might introduce a future bug if somebody tries to use the time value as a "real" time.

TimSmithCtx (approved):

PR #580, opened 2022-02-16
Fix timeout_call function

Wescoeur:
Ensure alarm is correctly reset in case of success after user function execution.
Otherwise the SIGALRM signal can be emitted after the execution of the given function.

### DIFF
--- a/drivers/util.py
+++ b/drivers/util.py
@@ -886,10 +886,9 @@ def timeout_call(timeoutseconds, function, *arguments):
     signal.signal(signal.SIGALRM, handler)
     signal.alarm(timeoutseconds)
     try:
-        function( * arguments)
-    except:
+        function(*arguments)
+    finally:
         signal.alarm(0)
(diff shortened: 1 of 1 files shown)


drivers/util.py
```diff
@@ -886,7 +886,9 @@ def handler(signum, frame):
     signal.signal(signal.SIGALRM, handler)
     signal.alarm(timeoutseconds)
     try:
-        function( * arguments)
+        result = function(*arguments)
+        signal.alarm(0)
```
edwintorok, line 890: If this needs to be executed on both success and failure cases, it might be nicer for this to go into a `finally:` section

Wescoeur: Right. I hesitated. :) I will repush!

rosslagerwall (approved):

TimSmithCtx (approved):

edwintorok (approved):

PR #582, opened 2022-03-18
Raise explicit error in case of NFS mount failure

Wescoeur:
During creation of ISOSR with NFS params, we can get a generic error with this command line:

```
xe sr-create host-uuid=<uuid> content-type=iso type=iso name-label=<loc> device-config:location=<invalid_server_path>
```

Before this patch the returned error was:

```
Error code: SR_BACKEND_FAILURE_1200
Error parameters: , ,
```

After:

```
Error code: SR_BACKEND_FAILURE_222
Error parameters: , Could not mount the directory specified in Device Configuration [opterr=mount failed with return code 32],
```

### DIFF
--- a/drivers/ISOSR.py
+++ b/drivers/ISOSR.py
@@ -364,6 +364,8 @@ class ISOSR(SR.SR):
             else:
                 raise xs_errors.XenError(
                     'ISOMountFailure', opterr=inst.reason)
+        except nfs.NfsException as e:
+            raise xs_errors.XenError('ISOMountFailure', opterr=str(e.errstr))
 
         # Check the iso_path is accessible
         if not self._checkmount():
--- a/drivers/nfs.py
+++ b/drivers/nfs.py
@@ -135,10 +135,12 @@ def soft_mount(mountpoint, remoteserver, remotepath, transport, useroptions='',
     # Wait for NFS service to be available
     try:
         if not check_server_service(remoteserver):
-            raise util.CommandException(code=errno.EOPNOTSUPP,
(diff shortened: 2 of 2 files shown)


edwintorok:
I think this is a good improvement.
Could this also catch the exception when an SMB connection is used for ISOs?
Might be nice to also include the device-config (or IP of server) in the error message (would make triaging failures easier if you can see which server in your infrastructure keeps raising errors).

Wescoeur:
It seems there is already a good exception handling with samba:
```
> xe sr-create host-uuid=<UUID> type=iso name-label=ISO-SR content-type=iso shared=true device-config:location="\\192.168.1.1\sambatest" device-config:username=xen device-config:cifspassword=<PASS>  device-config:type=cifs device-config:vers=3.0
Error code: SR_BACKEND_FAILURE_222
Error parameters: , Could not mount the directory specified in Device Configuration [opterr=mount.cifs: bad UNC (\192.168.1.1\sambatest)],
```

So I just updated the PR with small modifications to display the remote server in the NFS case. :slightly_smiling_face: 
Now we have:
```
> xe sr-create host-uuid=<UUID>  content-type=iso type=iso name-label="ISO SR" device-config:location=192.168.1.1:/mnt/xen/qa-slaves/iso
Error code: SR_BACKEND_FAILURE_222
Error parameters: , Could not mount the directory specified in Device Configuration [opterr=mount failed on server `192.168.1.1` with return code 32], 
```

MarkSymsCtx (approved):

PR #584, opened 2022-04-11
Added Dell ME4 multipath config

vafgoettlich:
Hi all,

we are using some Dell ME4 iSCSI-NAS Devices. Would be great to have a good multipath-config in place by default. For details: see https://downloads.dell.com/manuals/common/powervault-me4-series-linux-3924-bp-l_en-us.pdf

Regards,

Ulrich Göttlich

### DIFF
--- a/multipath/multipath.conf
+++ b/multipath/multipath.conf
@@ -41,6 +41,17 @@ devices {
 		no_path_retry		30
 	}
 	device {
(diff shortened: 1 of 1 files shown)


MarkSymsCtx (approved):
LGTM. Dell have approved it as well.

rosslagerwall (approved):

PR #586, opened 2022-06-13
Remove use of eval

MarkSymsCtx:

### DIFF
--- a/drivers/LVHDSR.py
+++ b/drivers/LVHDSR.py
@@ -1216,7 +1216,7 @@ class LVHDSR(SR.SR):
             if hostRef == masterRef:
                 continue
             util.SMlog("Deactivate VDI on %s" % hostRef)
-            rv = eval(self.session.xenapi.host.call_plugin(hostRef, self.PLUGIN_ON_SLAVE, "multi", args))
+            rv = self.session.xenapi.host.call_plugin(hostRef, self.PLUGIN_ON_SLAVE, "multi", args)
             util.SMlog("call-plugin returned: %s" % rv)
             if not rv:
                 raise Exception('plugin %s failed' % self.PLUGIN_ON_SLAVE)
@@ -1240,8 +1240,8 @@ class LVHDSR(SR.SR):
                 continue
             util.SMlog("Updating %s, %s, %s on slave %s" % \
                     (origOldLV, origLV, baseLV, hostRef))
-            rv = eval(self.session.xenapi.host.call_plugin(
-                    hostRef, self.PLUGIN_ON_SLAVE, "multi", args))
+            rv = self.session.xenapi.host.call_plugin(
+                hostRef, self.PLUGIN_ON_SLAVE, "multi", args)
             util.SMlog("call-plugin returned: %s" % rv)
             if not rv:
                 raise Exception('plugin %s failed' % self.PLUGIN_ON_SLAVE)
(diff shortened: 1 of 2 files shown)


rosslagerwall (approved):

rosslagerwall:
Why was it using `eval()` in the first place? It just seems unnecessary in this context.

drivers/LVHDSR.py
```diff
@@ -1216,7 +1216,7 @@ def _updateSlavesPreClone(self, hostRefs, origOldLV):
             if hostRef == masterRef:
                 continue
             util.SMlog("Deactivate VDI on %s" % hostRef)
-            rv = eval(self.session.xenapi.host.call_plugin(hostRef, self.PLUGIN_ON_SLAVE, "multi", args))
+            rv = self.session.xenapi.host.call_plugin(hostRef, self.PLUGIN_ON_SLAVE, "multi", args)
```
edwintorok, line 1219: This is calling the https://github.com/xapi-project/sm/blob/master/drivers/on_slave.py plugin, which seems to return the string true on success.
On failure it'll go into the wrapper code here https://github.com/xapi-project/xen-api/blob/master/scripts/examples/python/XenAPIPlugin.py where it seems that on failure you get an xmlrpc dump of the failure. That xml failure then gets wrapped into a generic api failure https://github.com/xapi-project/xen-api/blob/fa6f118fb512a2c90159368a1a0bc1cc730c895b/ocaml/xapi/xapi_plugins.ml. And those failures get translated into python exceptions:
```
>>> import XenAPI
>>> x=XenAPI.xapi_local()
>>> x.xenapi.login_with_password('','')
>>> h=x.xenapi.host.get_all()[0]
>>> x.xenapi.host.call_plugin(h, "on-slave", "multi", [])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 269, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 163, in xenapi_request
    result = _parse_result(getattr(self, methodname)(*full_params))
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 243, in _parse_result
    raise Failure(result['ErrorDescription'])
XenAPI.Failure: ['INTERNAL_ERROR', '(Failure "Expecting Rpc.Dict, but found \'[]\'")']
>>> x.xenapi.host.call_plugin(h, "on-slave", "multi", {})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 269, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 163, in xenapi_request
    result = _parse_result(getattr(self, methodname)(*full_params))
  File "/usr/lib/python2.7/site-packages/XenAPI.py", line 243, in _parse_result
    raise Failure(result['ErrorDescription'])
XenAPI.Failure: ['XENAPI_PLUGIN_FAILURE', 'multi', 'KeyError', "'vgName'"]
```

So the test below for checking for a failure when `not rv` is probably redundant: an exception would've been raised.
I think it was using eval before to convert the string True to the boolean true, but that seems unnecessary: if the plugin always raises exceptions then any non-empty string can be treated as a success.

MarkSymsCtx:
> Why was it using `eval()` in the first place? It just seems unnecessary in this context.

Indeed, hysterical raisons probably. SQ picked it up and as commented about it looks entirely redundant

PR #588, opened 2022-08-08
CA-369613: report errors correctly from multipath

MarkSymsCtx:

### DIFF
--- a/drivers/XE_SR_ERRORCODES.xml
+++ b/drivers/XE_SR_ERRORCODES.xml
@@ -778,6 +778,17 @@
                 <description>Device Mapper path missing</description>
                 <value>432</value>
         </code>
+
+        <code>
+                <name>MultipathDeviceNotAppeared</name>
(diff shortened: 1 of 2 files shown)


TimSmithCtx (approved):

edwintorok (approved):
These look OK. Are there any XenCenter updates needed as a result of this, or will the SDK automatically pick this up on next build?

edwintorok:
@kc284 FYI

MarkSymsCtx:
> These look OK. Are there any XenCenter updates needed as a result of this, or will the SDK automatically pick this up on next build?

Good point, yes we'll should add these as localised message although they should only occur in rare error cases, i.e. if entirely bogus input data is provided, it just happened that XenRT did that as a result of commissioning a new SAN.

kc284:
> These look OK. Are there any XenCenter updates needed as a result of this, or will the SDK automatically pick this up on next build?

It will pick them automatically (see first line of sdk recipe in xapi's Makefile).

PR #589, opened 2022-08-08
CA-369395: default multipath handle to dmp if not set

MarkSymsCtx:
If
```
xe host-param-set multipathing=true uuid=<uuid>
```
is used it doesn't set the multipathhandle (which XenCenter does do), but as we only support Device Mapper Multipath mutipathing now we can set that as a default if multipathing is enabled in this way.

This doesn't change the other path of reading from `dconf` which is a deprecated mechanism anyway as that requires `multipathhandle` to be supplied in order to activate the path.

### DIFF
--- a/drivers/SR.py
+++ b/drivers/SR.py
@@ -487,7 +487,7 @@ class SR(object):
             else:
                 hconf = self.session.xenapi.host.get_other_config(self.host_ref)
                 self.mpath = hconf['multipathing']
-                self.mpathhandle = hconf['multipathhandle']
+                self.mpathhandle = hconf.get('multipathhandle', 'dmp')
 
             if self.mpath != "true":
                 self.mpath = "false"


edwintorok (approved):
Looks good for a minimal change, although if we don't support other config values for multipathhandle, and we don't intend to add more in the future can we drop the code that checks and uses mpathhandle to match the codepath that mpathhandle == dmp would take?

MarkSymsCtx:
> Looks good for a minimal change, although if we don't support other config values for multipathhandle, and we don't intend to add more in the future can we drop the code that checks and uses mpathhandle to match the codepath that mpathhandle == dmp would take?

That's a much more invasive and dangerous change and would need a significant amount of validation testing to ensure that we don't suddenly end up with something trying to dereference something which hasn't been initialised.

edwintorok:
In this case the code to do this is in `def _mpathHandle` which could be changed to import `mpath_dmp` and call the activate/deactivate methods, and then remove all references to multipathhandle and mpathhandle from the codebase.
As the code is currently written it would seem to me it is slightly buggy and would call activate on `mpath_dmp` and deactivate on `mpath_null` (because mpath would be false, which leads to assigning a null handle), but I could be wrong.
Cleaning all that up should make the code simpler and more robust (I see your point about testing though, so perhaps it could be done in a followup PR).

MarkSymsCtx:
> In this case the code to do this is in `def _mpathHandle` which could be changed to import `mpath_dmp` and call the activate/deactivate methods, and then remove all references to multipathhandle and mpathhandle from the codebase. As the code is currently written it would seem to me it is slightly buggy and would call activate on `mpath_dmp` and deactivate on `mpath_null` (because mpath would be false, which leads to assigning a null handle), but I could be wrong. Cleaning all that up should make the code simpler and more robust (I see your point about testing though, so perhaps it could be done in a followup PR).

Not quite, there's all these references for sure and possibly more  knock on impacts

```
> grep -r  self\.mpathmodule drivers/*
drivers/BaseISCSI.py:                self.mpathmodule.refresh(self.dconf['SCSIid'], 0)
drivers/BaseISCSI.py:            self.mpathmodule.reset(self.dconf['SCSIid'], explicit_unmap=True)
drivers/BaseISCSI.py:            self.mpathmodule.reset(SCSIid, explicit_unmap=True)
drivers/BaseISCSI.py:        path = self.mpathmodule.path(SCSIid)
drivers/HBASR.py:            path = self.mpathmodule.path(scsiutil.getSCSIid(vdi_path))
drivers/LVHDoHBASR.py:            self.mpathmodule.refresh(self.SCSIid, 0)
drivers/LVHDoHBASR.py:                self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
drivers/LVHDoHBASR.py:            self.mpathmodule.refresh(self.SCSIid, 0)
drivers/LVHDoHBASR.py:                self.mpathmodule.refresh(self.SCSIid, 0)
drivers/LVHDoHBASR.py:                self.mpathmodule.refresh(self.SCSIid, 0)
drivers/LVHDoHBASR.py:            self.mpathmodule.refresh(self.SCSIid, 0)
drivers/LVHDoHBASR.py:                self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
drivers/LVHDoHBASR.py:                self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
drivers/LVHDoHBASR.py:        self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
drivers/LVHDoHBASR.py:                self.mpathmodule.reset(self.SCSIid, explicit_unmap=True)
drivers/SR.py:        self.mpathmodule = __import__(module_name)
drivers/SR.py:            self.mpathmodule.activate()
drivers/SR.py:            self.mpathmodule.deactivate()
drivers/SR.py:        self.dconf['device'] = self.mpathmodule.path(SCSIid)
```

plus fixing up all the unit tests etc.

edwintorok:
Perhaps `self.mpathmodule = __import__("mpath_dmp")` could be one step that doesn't have as much impact (and could still allow you to introduce more methods in the future again if desired, + dead code removal around mpathhandle if desired), but makes activate/deactivate handling less buggy?
AFAICT `deactivate` is never called currently on `mpath_dmp` (the one in `null` is a noop), is that correct? What happens if you activate and deactivate multipath on a host multiple times?

Although given that mpath_dmp.deactivate was probably never called it could as well be buggy/bitrotten itself.

MarkSymsCtx:
> Perhaps `self.mpathmodule = __import__("mpath_dmp")` could be one step that doesn't have as much impact (and could still allow you to introduce more methods in the future again if desired, + dead code removal around mpathhandle if desired), but makes activate/deactivate handling less buggy? AFAICT `deactivate` is never called currently on `mpath_dmp` (the one in `null` is a noop), is that correct? What happens if you activate and deactivate multipath on a host multiple times?
> 
> Although given that mpath_dmp.deactivate was probably never called it could as well be buggy/bitrotten itself.

Indeed, but that will be the subject of further investigation and doesn't change the context for this change whcih causes multipath to fail to enable if the CLI instructions in the docs are followed - https://docs.citrix.com/en-us/citrix-hypervisor/8-2/storage/multipath.html#configure-multipathing

TimSmithCtx (approved):

PR #591, opened 2022-08-23
Update test lvmlib to use argparse

MarkSymsCtx:

### DIFF
--- a/tests/lvmlib.py
+++ b/tests/lvmlib.py
@@ -1,8 +1,8 @@
-import optparse
+import argparse
 import sys
 
 
-class TestOptParse(optparse.OptionParser):
+class TestArgParse(argparse.ArgumentParser):
     """ Override the default exit and error functions so that they don't print
     to stderr during the tests
     """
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
Just a modernisation of this test support library which was getting in the way of doing some fault tracing.

tests/lvmlib.py
```diff
@@ -72,51 +72,53 @@ def fake_lvchange(self, args, stdin):
...
+        active = not args.inactive
+        assert args.zero in ['y', 'n']
+        zeroed = args.zero == 'y'
 
         self.get_volume_group(vgname).add_volume(
-            options.name,
-            int(options.size_mb),
-            options.tag,
+            args.name,
+            int(args.size_mb),
```
rosslagerwall, line 101: Not that it matters particularly but it might be preferable to `parser.add_argument("-L", dest='size_mb', type=int)` instead.

rosslagerwall (dismissed):

tests/lvmlib.py
MarkSymsCtx: Yes, sensible, it was a basic search replace for the most part but that would be a further improvement in maintainability here.

TimSmithCtx (approved):

edwintorok (dismissed):
This is all test code so I assume any failures would've shown up in the CI

MarkSymsCtx:
> This is all test code so I assume any failures would've shown up in the CI

Indeed, all test code is required to be 100% covered.

rosslagerwall (approved):

TimSmithCtx (approved):

PR #594, opened 2022-09-20
CA-370572: relinking is a transient property, do not copy to clones

MarkSymsCtx:


MarkSymsCtx:
This, very small, race window cropped up in the Long Haul 14 days storage soak test and prevented a newly created VDI clone from being used by a running VM (e.g. in the case of a VM clone from snapshot) if, at the time the clone operation happened, the VHD tree for the source was in the process of being relinked after a garbage collection operation.

rosslagerwall (approved):

TimSmithCtx (approved):

PR #595, opened 2022-09-21
CA-370696 Do not attempt to validate device or NFS server paths

TimSmithCtx:
If a device does not exist, then this is a separate error. If it does, then by definition its path is a valid one so the check is unnecessary and serves only to cause problems when new device has some unexpected (but valid) character in its name.

Likewise whether or not an NFS server path string is valid is the business of that NFS server, not the client code which is attempting to mount it.

Because checkroot() is removed (it has no remaining purpose), some subclasses of the SR class need to be modified as they relied on a side-effect of it being called.

### DIFF
--- a/drivers/EXTSR.py
+++ b/drivers/EXTSR.py
@@ -77,7 +77,7 @@ class EXTSR(FileSR.FileSR):
 
         # Check PVs match VG
         try:
-            for dev in self.root.split(','):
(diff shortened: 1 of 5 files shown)


MarkSymsCtx (approved):
LGTM

rosslagerwall (approved):

PR #599, opened 2022-11-16
CA-372772: fix miscalculation of seek offset

MarkSymsCtx:

### DIFF
--- a/drivers/util.py
+++ b/drivers/util.py
@@ -624,7 +624,7 @@ def zeroOut(path, fromByte, bytes):
 
     blocks = bytes / blockSize
     bytes -= blocks * blockSize
-    fromByte = fromBlock + blocks * blockSize
+    fromByte = (fromBlock + blocks) * blockSize
     if blocks:
         cmd = [CMD_DD, "if=/dev/zero", "of=%s" % path, "bs=%s" % blockSize, \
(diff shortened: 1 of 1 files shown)


drivers/util.py
```diff
@@ -624,7 +624,7 @@ def zeroOut(path, fromByte, bytes):
 
     blocks = bytes / blockSize
```
rosslagerwall, line 625: Does this code work properly in Python 3 since blocks may be a float which will throw off the calculations below?

```
$ python2 -c 'print 5000/4096'
1

$ python3 -c 'print(5000/4096)'
1.220703125
```

Use `//` if you want the old Python 2 behaviour.

rosslagerwall (approved):

drivers/util.py
MarkSymsCtx: good spot, I'll catch that up in the Python3 migration changes.

MarkSymsCtx:
> I think this function would be a lot simpler if it were replaced with a pure Python implementation (and wouldn't require spawning up to 3 dd processes). It's not even using O_DIRECT so the precise block size doesn't really matter since the page cache will take care of that.

As a future enhancement, yes agreed.

TimSmithCtx (approved):

PR #600, opened 2023-01-27
Add warning to multipath.conf and provide new custom.conf for users to edit

stormi:
Users must not modify /etc/multipath.conf directly (despite various documentation asking them to do so, which should be fixed), because the file gets overwritten from time to time, when the defaults are updated. Instead, they should add their custom configuration into /etc/multipath/conf.d/.

This PR adds a warning at the top of multipath.conf and also a new custom.conf file for users to modify, with some onboarding contents


multipath/multipath.conf
```diff
@@ -1,3 +1,11 @@
+# --- WARNING: DO NOT EDIT THIS FILE ---
+# The contents of this file may be overwritten at any future time through a
+# system update, causing any custom configuration to be lost.
+#
+# For custom multipath configuration, create a separate .conf file in the
+# /etc/multipath/conf.d/ directory.
```
stormi, line 6: To be changed into an invitation to modify the /etc/multipath/conf.d/custom.conf if my proposal consisting in installing such a file by default on the system is accepted.

multipath/multipath.conf
```diff
@@ -1,3 +1,11 @@
...
+# --- END OF WARNING ---
```
MarkSymsCtx, line 7: Maybe include the `Refer to "man multipath.conf"` bit here as well?

multipath/custom.conf
```diff
@@ -0,0 +1,6 @@
+# Custom configuration for multipathd
+
+# Changes made to this file will not be overwritten by future system updates.
+# They will also be retained through system upgrades to newer releases.
+
+# Refer to "man multipath.conf"
```
stormi, line 6: Ideally, this file should contain a commented out skeleton, for easier customization. I don't master multipath configuration, so I haven't done it, but I'm sure @MarkSymsCtx, @Fohdeesha, or someone else will be able to come up with something if my idea is accepted.

MarkSymsCtx (changes requested):
Minor tweak I think and then it's good but I'll hold on merging it until we've finished all the validation work on the Python2->Python3 migration which is currently in progress.

multipath/custom.conf
```diff
@@ -0,0 +1,6 @@
+# Custom configuration for multipathd
+
+# Changes made to this file will not be overwritten by future system updates.
+# They will also be retained through system upgrades to newer releases.
```
stormi, line 4: This is not true at the moment, but I suggest we make it become true. If this is seen as a good idea, we can probably contribute the change to the installer, unless there are important technical challenges.

multipath/custom.conf
```diff
@@ -0,0 +1,6 @@
+# Custom configuration for multipathd
```
stormi, line 1: My suggestion with this file is to add it by default on systems and tell users about it in the main multipath.conf (as well as documentation), so that they have an easy entry point for custom configuration. The file would be marked `%config(noreplace)` in the spec file so that updates don't overwrite user changes in it.

multipath/multipath.conf
stormi: Wouldn't this be a repetition of what line 11 already says?

If we replace the message with something telling users to modify custom.conf, then this will not be needed anymore. 

Else, maybe we can move line 11 up.

stormi:
Would it be the right time to resume work on this PR?

multipath/multipath.conf
MarkSymsCtx: I think it's better for vendor neutrality for the configuration overrides to go into separate files.

stormi: I'm not sure about what you mean. Since the PR also provides a ready-to-modify file to be placed in `/etc/multipath/conf.d/custom.conf `, wouldn't it be appropriate to mention it?

PxDave:
So does that mean the vendors in the defaults file are removed?  What is neutral?

MarkSymsCtx:
> So does that mean the vendors in the defaults file are removed? What is neutral?

No, it means that we issue guidance that if vendors need to recommend additional or alternative configuration then it should go into a vendor specific file in /etc/multipath/conf.d and not just mixing all the bits into a single file. This means that it is possible for vendors to provide a complete, drop-in, file to add to the customer's hyypervisor deployment to enable the multipath configuration.

It is always preferable, but not necessarily possible, for vendors to send a pull-request to this file to have their preferred configuration installed by default and not requiring their (and our) customers to make modifications to the system which always have a risk (however small) of introducing a failure mode.

TimSmithCtx (approved):

PR #605, opened 2023-03-07
Remove obsolete unused pip confiuration

MarkSymsCtx:


rosslagerwall (approved):

edwintorok:
This is now failing due to the newer version of pylint complaining.

MarkSymsCtx:
> This is now failing due to the newer version of pylint complaining.

Yes, and as Github has a different version of pylint3 to what we have it can't be striaghtforwardly fixed without repeated pushes to github :(

ydirson:
> Yes, and as Github has a different version of pylint3 to what we have it can't be striaghtforwardly fixed without repeated pushes to github :(

I can reproduce quite easily those errors by installing in a venv.

ydirson:
If that's OK with you @MarkSymsCtx I'll deal with the pylint errors to unblock this PR

PR #606, opened 2023-03-07
Pylint fixes

ydirson:
This is a first quick pass on errors raised by pylint 2.16.4. Nothing critical, mostly cleanups that also improve readability (and as a side effect can help in the future to enable type checking).

The goal is to make #605 mergeable - we're not there yet, but a few similar fixes should do.

### DIFF
--- a/drivers/srmetadata.py
+++ b/drivers/srmetadata.py
@@ -652,25 +652,24 @@ class LVMMetadataHandler(MetadataHandler):
     def spaceAvailableForVdis(self, count):
         created = False
         try:
-            try:
-                # The easiest way to do this, is to create a dummy vdi and write it
-                uuid = util.gen_uuid()
-                vdi_info = {UUID_TAG: uuid,
-                            NAME_LABEL_TAG: 'dummy vdi for space check',
-                            NAME_DESCRIPTION_TAG: 'dummy vdi for space check',
-                            IS_A_SNAPSHOT_TAG: 0,
-                            SNAPSHOT_OF_TAG: '',
-                            SNAPSHOT_TIME_TAG: '',
-                            TYPE_TAG: 'user',
-                            VDI_TYPE_TAG: 'vhd',
-                            READ_ONLY_TAG: 0,
-                            MANAGED_TAG: 0,
-                            'metadata_of_pool': ''
-                }
+            # The easiest way to do this, is to create a dummy vdi and write it
+            uuid = util.gen_uuid()
+            vdi_info = {UUID_TAG: uuid,
+                        NAME_LABEL_TAG: 'dummy vdi for space check',
+                        NAME_DESCRIPTION_TAG: 'dummy vdi for space check',
+                        IS_A_SNAPSHOT_TAG: 0,
+                        SNAPSHOT_OF_TAG: '',
+                        SNAPSHOT_TIME_TAG: '',
+                        TYPE_TAG: 'user',
+                        VDI_TYPE_TAG: 'vhd',
+                        READ_ONLY_TAG: 0,
+                        MANAGED_TAG: 0,
+                        'metadata_of_pool': ''
+            }
 
-                created = self.addVdiInternal(vdi_info)
-            except IOError as e:
-                raise
+            created = self.addVdiInternal(vdi_info)
+        except IOError as e:
+            raise
         finally:
             if created:
                 # Now delete the dummy VDI created above
--- a/Makefile
+++ b/Makefile
@@ -110,15 +110,7 @@ precommit: build
 
 .PHONY: precheck
 precheck: build
-	@ QUIT=0; \
-	for i in $(SM_PY_FILES); do \
-		echo Checking $${i} ...; \
-		PYTHONPATH=./drivers:$$PYTHONPATH $(PYLINT) --rcfile=tests/pylintrc $${i}; \
-		[ $$? -ne 0 ] && QUIT=1 ; \
-	done; \
-	if [ $$QUIT -ne 0 ]; then \
-		exit 1; \
-	fi; \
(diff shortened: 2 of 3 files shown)


MarkSymsCtx:
Looks like we might want some combination of this and https://github.com/xapi-project/sm/pull/607 (which is the recreation of #605 having got it to actually complete the actions properly)

ydirson:
Nice!  How should we handle this?  #607 is more complete, it could make sense that you pick the things you like from this one and we'll close it?

MarkSymsCtx:
> Nice! How should we handle this? #607 is more complete, it could make sense that you pick the things you like from this one and we'll close it?

Now that #607 is merged rebase this and then we're re-review it.

drivers/VDI.py
```diff
@@ -858,13 +858,13 @@ def _create_cbt_log(self):
 
         return logpath
 
-    def _activate_cbt_log(self, logname):
+    def _activate_cbt_log(self, logname) -> bool:
```
MarkSymsCtx, line 861: We don't, currently at least, have typing enabled so this is going to have much affect but should be innocuous.

drivers/srmetadata.py
```diff
@@ -633,11 +633,11 @@ def getMetadataToWrite(self, sr_info, vdi_info, lower, upper, update_map, \
             raise
 
     # specific functions, to be implement by the child classes
-    def getVdiInfo(self, Dict, generateSector=0):
-        return
+    def getVdiInfo(self, Dict, generateSector=0) -> str:
+        raise NotImplementedError()
```
MarkSymsCtx, line 637: This will cause failures, the correct resolution was in #607 so just drop this and the similar change in `getSRInfoForSectors`

drivers/util.py
```diff
@@ -1228,8 +1228,10 @@ def daemon():
 #########################
 
 if __debug__:
+    # FIXME this needs some explanation as to why os.environ.get() is
+    # not called unconditionally
     try:
-        XE_IOFI_IORETRY
+        XE_IOFI_IORETRY         # pylint: disable=used-before-assignment
```
MarkSymsCtx, line 1234: This whole thing looks truly weird, it was weird before so it's not a reflection of this change but I wonder whether we might not be better seeing if this is actually even used and removing it if not.

MarkSymsCtx (changes requested):

drivers/util.py
MarkSymsCtx: This does indeed seem to be entirely unused and legacy cruft so please just delete the entire `if __debug__:` block

ydirson:
While I was at it I tackled as much as the remaining pylint 2.13.9 errors are possible.  A few of them are not so obviously fixed, and I flagged those commits as WIP for discussion.

drivers/LVHDoISCSISR.py
```diff
@@ -492,6 +492,9 @@ def delete(self, sr_uuid):
     def attach(self, sr_uuid):
         try:
             connected = False
+            if not self.iscsiSRs:
```
MarkSymsCtx, line 495: instead of `if not self.iscsiSRs:` you might  have better results with `if len(self.iscsiSRs) == 0:`.

It could also be confused by the `try: except: else:`  construct to set `connected=True`, that could be moved into the main block and the `else:` removed.

MarkSymsCtx:
> While I was at it I tackled as much as the remaining pylint 2.13.9 errors are possible. A few of them are not so obviously fixed, and I flagged those commits as WIP for discussion.

I think (but will need to dig a bit further` that  `_genArrayIdentifier` is entirely unused dead code as it's only called under conditions relating to MPP multipath which is no longer used and we now only use device-mapper-multipath for multipath devices. Expunging all of that dead code could be a chunk of work.

ydirson:
> > While I was at it I tackled as much as the remaining pylint 2.13.9 errors are possible. A few of them are not so obviously fixed, and I flagged those commits as WIP for discussion.
> 
> I think (but will need to dig a bit further`that `_genArrayIdentifier` is entirely unused dead code as it's only called under conditions relating to MPP multipath which is no longer used and we now only use device-mapper-multipath for multipath devices. Expunging all of that dead code could be a chunk of work.

You mean that's better left for a separate PR ?  I guess that would amount to reworking the `elif obj.hba == 'mpp':`  block to warn as ba1338442e84f6906a52ae8481ada39f08abb48a does, and select some fallback behavior so we can get rid of the dead code, but I don't fee confident enough to propose said fallback behavior :)

MarkSymsCtx:
> > > While I was at it I tackled as much as the remaining pylint 2.13.9 errors are possible. A few of them are not so obviously fixed, and I flagged those commits as WIP for discussion.
> > 
> > 
> > I think (but will need to dig a bit further`that `_genArrayIdentifier` is entirely unused dead code as it's only called under conditions relating to MPP multipath which is no longer used and we now only use device-mapper-multipath for multipath devices. Expunging all of that dead code could be a chunk of work.
> 
> You mean that's better left for a separate PR ? I guess that would amount to reworking the `elif obj.hba == 'mpp':` block to warn as [ba13384](https://github.com/xapi-project/sm/commit/ba1338442e84f6906a52ae8481ada39f08abb48a) does, and select some fallback behavior so we can get rid of the dead code, but I don't fee confident enough to propose said fallback behavior :)

Yeah, I'm doing it now. About to run some tests to make sure nothing explodes in multipath testing, it all seemed to be dangling dead code with no way of triggering the relevant code paths. 207 lines of dead code removed.

MarkSymsCtx:
@ydirson FYI - https://github.com/xapi-project/sm/pull/613

MarkSymsCtx (approved):

TimSmithCtx (approved):

MarkSymsCtx (approved):

rdn32 (approved):

PR #608, opened 2023-03-08
Fix under-quoted backslashes by using raw strings

ydirson:
Running tests would show such warnings:

```
tests/test_ISOSR.py:201
  /data/src/sm/tests/test_ISOSR.py:201: DeprecationWarning: invalid escape sequence \j
    username='citrix\jsmith',

drivers/ISOSR.py:59
  /data/src/sm/drivers/ISOSR.py:59: DeprecationWarning: invalid escape sequence \.
    regex = re.compile("\.iso$|\.img$", re.I)
```

### DIFF
--- a/tests/test_ISOSR.py
+++ b/tests/test_ISOSR.py
@@ -201,7 +201,7 @@ class TestISOSR_overSMB(unittest.TestCase):
         context.setup_error_codes()
         update = {'cifspassword': 'winter2019'}
         smbsr = self.create_smbisosr(atype='cifs', vers='1.0',
-                                     username='citrix\jsmith',
+                                     username=r'citrix\jsmith',
                                      dconf_update=update)
         _checkmount.side_effect = [False, True]
         smbsr.attach(None)
--- a/drivers/ISOSR.py
+++ b/drivers/ISOSR.py
@@ -56,7 +56,7 @@ NFSPORT = 2049
 
 def is_image_utf8_compatible(s):
     # pylint: disable=no-member
-    regex = re.compile("\.iso$|\.img$", re.I)
+    regex = re.compile(r"\.iso$|\.img$", re.I)
     if regex.search(s) is None:
         return False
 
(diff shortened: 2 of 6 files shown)


tests/test_ISOSR.py
```diff
@@ -198,7 +198,7 @@ def test_attach_with_smb_credentials_domain(self, context,
         context.setup_error_codes()
         update = {'cifspassword': 'winter2019'}
         smbsr = self.create_smbisosr(atype='cifs', vers='1.0',
-                                     username='citrix\jsmith',
+                                     username=r'citrix\jsmith',
```
MarkSymsCtx, line 201: this is not strictly a correct use of an `r` string, it should really just be either `citrix\\jsmith` or drop the domain piece entirely

MarkSymsCtx (changes requested):

tests/test_ISOSR.py
ydirson: Why so ? Raw string literals are most commonly used for regexps, but there's no reason I know of to avoid them in other contexts.  Although in the python standard library doc there are only mentions in the regexp context, [the language reference](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) has no such restriction.

MarkSymsCtx: OK, fair enough. I was only ever introduced to these as being a regular expression declarator.

andyhhp: `r` means raw, not regex.  Regexes tend to contain a lot of `\` which is why it is common for them to be written as raw strings, but they're just strings
```python
>>> type(r"")
<class 'str'>
```

MarkSymsCtx (approved):

TimSmithCtx (approved):

MarkSymsCtx:
@ydirson probably needs a rebase now that #607 is merged.

MarkSymsCtx (approved):

TimSmithCtx (approved):

PR #616, opened 2023-04-13
FileSR: fix error code

ydirson:
The error code itself is short-lived and not used, but the "524" string is checked in various tests, so sync accordingly.

I admit I'm puzzled by the fact this check only happens when a "fist point" is active, I would have thought this check for hardlink support would be more generally useful than that?

### DIFF
--- a/drivers/FileSR.py
+++ b/drivers/FileSR.py
@@ -1071,7 +1071,7 @@ class SharedFileSR(FileSR):
     NO_HARDLINK_SUPPORT = "no_hardlinks"
 
     def _raise_hardlink_error(self):
-        raise OSError(542, "Unknown error 524")
+        raise OSError(524, "Unknown error 524")
 
     def _check_hardlinks(self):
         test_name = os.path.join(self.path, str(uuid4()))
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
> The error code itself is short-lived and not used, but the "524" string is checked in various tests, so sync accordingly.
> 
> I admit I'm puzzled by the fact this check only happens when a "fist point" is active, I would have thought this check for hardlink support would be more generally useful than that?

You're misinterpreting the usage. The check for hardlinks is, as you say, generically useful. The FIST point allows the test automation systems to force a failure in the hardlink check even in the absence of filers which do not support hardlinks (i.e. we don't have any in the test automation lab but at least one customer does), and thus allow testing the fallback code. I suspect the error may be that the string is wrong and the numeric is correct. Let me check the original issue details.

MarkSymsCtx (dismissed):

ydirson:
> The FIST point allows the test automation systems to force a failure

That makes sense, thanks for the clarification. Modified the comment to convey that info.

MarkSymsCtx (approved):

TimSmithCtx (approved):

Issue #551, opened 2021-04-15
Support NFSv4+ only storage repositories


jcharaoui:
On servers where NFSv3 is disabled and NFSv4+ is enabled and working, `rpcbind` and `showmount` commands do not work, causing the `nfs.py` helper functions to falsely conclude that the server is unreachable.

When activating an SR configured for NFSv4+, `check_server_tcp()` and `check_server_service()` should probably be skipped, or maybe execute a simple check that `2049/tcp` is open on the remote.

olivierlambert:
@edwintorok is there anything we can do to assist on this?

MarkSymsCtx:
There needs to be some check of connectivity viability otherwise incorrect details given to a probe operation will take a full TCP session timeout to fail (~75 seconds) and that's terrible user experience. Using util.testHost to hit 2049/tcp is probably a reasonable start for that. The showmount call I think can be replaced by connecting to the server with no share name and running an `ls` on it but whether this is universal or just something that the Linux  nfs-server-common does I'm not sure.

olivierlambert:
> TCP session timeout to fail (~75 seconds) and that's terrible user experience

I agree. So it's a 2 stages check:
1. TCP 2049 port check with `util.testHost`
2. Checking if we can actually mount it and browse it (I'm just no sure exactly what's the best way to do it, we might check what `nfs-server-common` is doing in the first place, as you suggested.

@marcungeschikts we need to get that in our backlog

emanzx:
I have been waiting for this issue to be fix for almost a year. I dont know how to help. But from what @MarkSymsCtx suggested I think that is ok. and I am not sure if this correct. I just try asking chatGPT about how mount verify the validity of the shared NFS and it gave me this answer.. 

> The mount command verifies the validity of a shared NFS file system in several ways. First, it performs a network-based check to see if the NFS server is reachable and if the specified NFS share is available on the server. The mount command uses the Remote Procedure Call (RPC) protocol to communicate with the NFS server and verify the availability of the share. The mount command can also check the validity of the NFS share by checking the file system type, and verifying that the NFS version and transport protocol are supported. Once the mount command has verified the validity of the NFS share, it will mount the share to the specified local mount point. If the NFS share is not available or if there is an error in the configuration, the mount command will return an error and the file system will not be mounted.

but when I asked NSFv4 above does not have RPC available and how mount verifies it.. I get this answer.

[...]

PR #617, opened 2023-05-02
Support NFSv4 only environment

benjamreis:
NFSv4 only environments don't support `rpcinfo`
and `showmount` calls as they're missing NFSv3
services.

When `rpcinfo` or `showmount` fails, try to mount
NFSv4 pseudo FS '/' to add '4' to supported NFS versions
and run `ls` on the mounted pseudo FS to offer the first
folder level of exports.

See: https://github.com/xapi-project/sm/issues/551
And: https://github.com/xcp-ng/xcp/issues/135

### DIFF
--- a/drivers/nfs.py
+++ b/drivers/nfs.py
@@ -182,6 +182,70 @@ def unmount(mountpoint, rmmountpoint):
             raise NfsException("rmdir failed with error '%s'" % inst.strerror)
 
 
+def _scan_exports_nfs3(target, dom, element):
+    """ Scan target and return an XML DOM with target, path and accesslist.
+        Using NFS3 services.
+    """
+
+    cmd = [SHOWMOUNT_BIN, "--no-headers", "-e", target]
+    for val in util.pread2(cmd).split('\n'):
+        if not len(val):
+            continue
+        entry = dom.createElement('Export')
+        element.appendChild(entry)
+
+        subentry = dom.createElement("Target")
+        entry.appendChild(subentry)
+        textnode = dom.createTextNode(target)
+        subentry.appendChild(textnode)
+
+        # Access is not always provided by showmount return
+        # If none is provided we need to assume "*"
+        array = val.split()
+        path = array[0]
+        access = array[1] if len(array) >= 2 else "*"
+        subentry = dom.createElement("Path")
+        entry.appendChild(subentry)
+        textnode = dom.createTextNode(path)
+        subentry.appendChild(textnode)
+
+        subentry = dom.createElement("Accesslist")
+        entry.appendChild(subentry)
+        textnode = dom.createTextNode(access)
+        subentry.appendChild(textnode)
+    return dom
+
+
+def _scan_exports_nfs4_only(target, transport, dom, element):
+    """ Scan target and return an XML DOM with target, path and accesslist.
+        Using NFS4 only pseudo FS.
+    """
+
+    mountpoint = "%s/%s" % (NFS4_TMP_MOUNTPOINT, target)
+    soft_mount(mountpoint, target, NSFv4_PSEUDOFS, transport, nfsversion='4')
+    paths = os.listdir(mountpoint)
+    unmount(mountpoint, NSFv4_PSEUDOFS)
+    for path in paths:
+        entry = dom.createElement('Export')
+        element.appendChild(entry)
+
+        subentry = dom.createElement("Target")
+        entry.appendChild(subentry)
+        textnode = dom.createTextNode(target)
+        subentry.appendChild(textnode)
+        subentry = dom.createElement("Path")
+        entry.appendChild(subentry)
+        textnode = dom.createTextNode(path)
+        subentry.appendChild(textnode)
+
+        subentry = dom.createElement("Accesslist")
+        entry.appendChild(subentry)
+        # Assume everyone as we do not have any info about it
+        textnode = dom.createTextNode("*")
+        subentry.appendChild(textnode)
+    return dom
+
+
 def scan_exports(target, transport):
     """Scan target and return an XML DOM with target, path and accesslist."""
     util.SMlog("scanning")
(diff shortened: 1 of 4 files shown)


benjamreis:
What can be done next:

Attempt a mount on / with nfs4 and:
- run `nfsstat -m` to get the version of the mount and add it to the supported version
- run ls in the mounted / to get some exported paths BUT what if there's multiple level of folders?

What do you think @MarkSymsCtx?

MarkSymsCtx (dismissed):

MarkSymsCtx:
> What can be done next:
> 
> Attempt a mount on / with nfs4 and:
> 
> * run `nfsstat -m` to get the version of the mount and add it to the supported version
> * run ls in the mounted / to get some exported paths BUT what if there's multiple level of folders?
> 
> What do you think @MarkSymsCtx?

It's all generally horrible with NFSv4 but the above may help in a significant set of sub-cases.

benjamreis:
Okay, i'll try that then, it's still better than nothing.
Would you prefer to merge this PR 1st and I'll do a 2nd one on top or wait and do everything in this PR?

stormi:
> run nfsstat -m to get the version of the mount and add it to the supported version 

I don't know exactly how this works, but the current version of `mount.nfs` in XS/XCP-ng dom0 defaulted to 4.1 for a server that only supports 4.1 and 4.2. Another client on a more recent computer defaulted to 4.2 for the same server.

This means that the version we can extract from `nfsstat -m` depends on the client and doesn't give much information about what the server supports, so maybe `probe` should just reply `4` as the supported version when there's no NFS v3 (Option A). 

Another approach would be, after a first successful mount of `/`, to attempt to mount `/` with every known version of the v4 protocol (4.0, 4.1, 4.2), and keep only the working values (Option B).

Yet another approach would be always listing all version of NFSv4, and let it fail at `sr-create` time when the user attempts a not-supported version (Option C).

drivers/nfs.py
```diff
@@ -282,11 +318,26 @@ def get_supported_nfs_versions(server):
...
+        mountpoint = "%s/%s" % NFS4_TMP_MOUNTPOINT, server
+        soft_mount(mountpoint, server, NSFv4_PSEUDOFS, transport, nfsversion='4')
+        util.pread2([NFS_STAT, '-m'])
+        unmount(mountpoint, NSFv4_PSEUDOFS)
+        if res is None:
+            res = ['4']
+        else:
+            res.add('4')
+    except Exception:
+        util.SMlog("Unable to obtain list of valid nfs versions with NSFv4 pseudo FS mount")
```
stormi, line 335: Is this needed at all when res is not None? IIRC, when the server supports both v3 and v4, rpcinfo already lists 4 among the versions, so there's no need to try to mount the pseudo filesystem from NFSv4 in this situation.

drivers/nfs.py
```diff
@@ -186,38 +194,74 @@ def unmount(mountpoint, rmmountpoint):
...
+            subentry.appendChild(textnode)
+
+            subentry = dom.createElement("Accesslist")
+            entry.appendChild(subentry)
+            # Assume everyone as we do not have any info about it
+            textnode = dom.createTextNode("*")
+            subentry.appendChild(textnode)
+        fail = False
+    except Exception:
+        util.SMlog("Unable to scan exports with NFSv4 pseudo FS mount")
```
stormi, line 260: AFAIK you only need to do this if showmount doesn't give any results. Otherwise, you already have the exports for both v3 and v4.

drivers/nfs.py
```diff
@@ -186,38 +194,74 @@ def unmount(mountpoint, rmmountpoint):
...
+            textnode = dom.createTextNode(path)
+            subentry.appendChild(textnode)
+
+            subentry = dom.createElement("Accesslist")
+            entry.appendChild(subentry)
+            textnode = dom.createTextNode(access)
+            subentry.appendChild(textnode)
+        fail = False
+    except Exception:
+        util.SMlog("Unable to scan exports with rpcingo, trying NFSv4")
```
stormi, line 233:
```suggestion
        util.SMlog("Unable to scan exports with %s, trying NFSv4" % SHOWMOUNT_BIN)
```

drivers/nfs.py
```diff
@@ -50,34 +52,40 @@
...
+def check_server_service(server, transport):
     """Ensure NFS service is up and available on the remote server.
 
     Returns False if fails to detect service after 
     NFS_SERVICE_RETRY * NFS_SERVICE_WAIT
     """
 
+    sv = get_supported_nfs_versions(server, transport)
+    if sv == ['4']:
+        return True
```
stormi, line 87: Maybe add a comment explaining why?

drivers/nfs.py
```diff
@@ -186,38 +194,74 @@ def unmount(mountpoint, rmmountpoint):
...
+        fail = False
+    except Exception:
+        util.SMlog("Unable to scan exports with rpcingo, trying NFSv4")
 
-        subentry = dom.createElement("Accesslist")
-        entry.appendChild(subentry)
-        textnode = dom.createTextNode(access)
-        subentry.appendChild(textnode)
+    try:
+        mountpoint = "%s/%s" % NFS4_TMP_MOUNTPOINT, target
```
stormi, line 236: Does this work without enclosing the tuple in parentheses?

```
>>> a = "%s %s" % "one", "two"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: not enough arguments for format string
```

benjamreis:
Successfull tests on NFS3 & 4.

olivierlambert:
Hey :) Any news on this? Is there anything left to do or simply getting validated?

drivers/nfs.py
```diff
@@ -186,43 +182,75 @@ def unmount(mountpoint, rmmountpoint):
...
-        subentry = dom.createElement("Accesslist")
-        entry.appendChild(subentry)
-        textnode = dom.createTextNode(access)
-        subentry.appendChild(textnode)
-
-    return dom
-
-
-def scan_srlist(path, dconf):
+    try:
```
MarkSymsCtx, line 191: I don't particularly like how long this method has now become, it makes it quite cognitively hard to follow. Would be good to pull some of these long open coded blocks out into well named methods which describe what they're doing and then each one can be reviewed and understood in isolation (or ignored if not of interest to the current needs).

MarkSymsCtx: Similarly, to a lesser extent, `get_supported_nfs_versions` below.

MarkSymsCtx (changes requested):

drivers/nfs.py
benjamreis: I can split in methods for NFS3 and NFS4Only.

benjamreis: Done :)

drivers/nfs.py
```diff
@@ -218,11 +213,61 @@ def scan_exports(target):
         entry.appendChild(subentry)
         textnode = dom.createTextNode(access)
         subentry.appendChild(textnode)
+    return dom
+
+
+def _scan_exports_nfs4_only(target, transport, dom, element):
```
MarkSymsCtx, line 219: given that we have `_scan_exports_nfs3` this should probably just be `_scan_exports_nfs4` for consistency.

MarkSymsCtx (commented):
Just a minor nit on method names.

Is there anything "useful" to see in the logging for this that will show it working?

drivers/nfs.py
benjamreis: I hesitated to add the `_only` or not, both are okay for me so I can remove it if you prefer.

benjamreis:
> Is there anything "useful" to see in the logging for this that will show it working?

In a NFS3/NFS3-4 env you should see the same logs as before since NFS4 only is a fallback in case of failure.
In NFS4 only env I added logs saying the 1st method fails and we're trying using the NFS4 pseudo FS
https://github.com/xapi-project/sm/pull/617/files#diff-9ca8a7292c88a517d48db9a9d25ee2d14ffd83f7f40d818c1c4121ce2e275b35R258
https://github.com/xapi-project/sm/pull/617/files#diff-9ca8a7292c88a517d48db9a9d25ee2d14ffd83f7f40d818c1c4121ce2e275b35R346

I also added logs in case both methods failed before throwing an exception.

drivers/nfs.py
benjamreis: done :)

drivers/nfs.py
```diff
@@ -50,34 +51,41 @@
...
+NFS4_TMP_MOUNTPOINT = "/tmp/mnt"
 
 class NfsException(Exception):
 
     def __init__(self, errstr):
         self.errstr = errstr
 
-
-def check_server_tcp(server, nfsversion=DEFAULT_NFSVERSION):
+def check_server_tcp(server, transport, nfsversion=DEFAULT_NFSVERSION):
```
MarkSymsCtx, line 62: The method is explicit (or is expected to be) in using TCP, so it should not have the transport argument. If `get_suppoted_nfs_versions` needs a transport it should be passed as a literal from this method without changing the signature of the method.

MarkSymsCtx (changes requested):

drivers/nfs.py
benjamreis: Indeed but transport can be `tcp` or `tcp6` in the case of IPv6.

MarkSymsCtx: Ah, I see, that was not clear. Might have been better for the IPv6 bits to be at least a commit on their own right to make it clear.

MarkSymsCtx (approved):

stormi (approved):

MarkSymsCtx:
@olivierlambert I'd already added the extra reviewer

KevinLCtx (approved):

olivierlambert:
Can you merge? Thanks!

KevinLCtx:
> Can you merge? Thanks!

Mark is on vacation I will see if I can find somebody else who has permission to merge it.

PR #619, opened 2023-05-16
CA-376740: smb: Fix snapshot with newer kernels

rosslagerwall:
Newer kernels defer closes by default but this interferes with use of hard links since with SMB (or maybe the underlying server filesystem), a hard link cannot be opened if the target is also open, which it may be temporarily with deferred close. Disable using the mount param available since Linux 6.0.

This can be reproduced with the following Python snippet. When deferred close is enabled, the final open fails with EINVAL.

    fd = os.open('test.vhd', os.O_WRONLY|os.O_CREAT)
    os.write(fd, b'foo')
    os.close(fd)

    fd = os.open('test.vhd', os.O_RDONLY|os.O_DIRECT)
    os.close(fd)

    os.link('test.vhd', 'new.vhd')
    newfd = os.open('new.vhd', os.O_RDONLY|os.O_DIRECT)

### DIFF
--- a/drivers/SMBSR.py
+++ b/drivers/SMBSR.py
@@ -155,6 +155,15 @@ class SMBSR(FileSR.SharedFileSR):
                 'actimeo=0'
         ]
 
+	# Newer kernels defer closes by default but this interferes with use of
+	# hard links since with SMB (or maybe the underlying server
+	# filesystem), a hard link cannot be opened if the target is also
+	# open (which it will be with deferred close). Disable using the
+	# mount param, available since Linux 6.0.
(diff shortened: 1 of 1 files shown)


MarkSymsCtx (approved):

rosslagerwall:
After discussion with upstream (https://lore.kernel.org/linux-cifs/DM6PR03MB537265393AE53D1A34F1F5CBF0409@DM6PR03MB5372.namprd03.prod.outlook.com/T/#t), this appears to be a kernel bug so I'll rather put a workaround into the kernel instead until it is fixed.

PR #622, opened 2023-06-20
CA-378768: set scheduler on multipath device

MarkSymsCtx:


drivers/mpath_dmp.py
```diff
@@ -49,6 +49,7 @@ def activate_MPdev(sid, dst):
     path = os.path.join(MP_INUSEDIR, sid)
     cmd = ['ln', '-sf', dst, path]
     util.pread2(cmd)
+    util.set_scheduler(os.path.realpath(path))
```
edwintorok, line 52: this sets the scheduler of the toplevel multipath mapper. Do we also need to set the scheduler for the underlying devices (if they are not shared with Dom0), or is that already done elsewhere?

MarkSymsCtx: That was already done/

edwintorok (approved):

PR #627, opened 2023-07-19
CA-379329: check for missing iSCSI sessions and reconnect

MarkSymsCtx:
If, when an iSCSI SR is attached, some of the paths are not reachable they will not be subsequently connected should the connectivity issues be resolved, without detaching the SR or rebooting the host. Neither of these options give a good experience for users of the software. Add a systemd timer which will, periodically (default 10 minutes), ask each SR to do SR specific "health" checks. Currently only a checker for the iSCSI sessions is present but more can be added to this pattern in the future as needs arise.

### DIFF
--- a/drivers/sr_health_check.py
+++ b/drivers/sr_health_check.py
@@ -1,0 +1,60 @@
+#!/usr/bin/python3
+
+# Copyright (C) Cloud Software Group, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; version 2.1 only.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+"""
+Health check for SR, to be triggered periodically by a systemd timer. What is checked is
+SR implementation type dependent.
+"""
+
+import SR
+import util
+
+
+def main():
+    """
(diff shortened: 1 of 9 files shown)


drivers/sr_health_check.py
```diff
@@ -0,0 +1,66 @@
...
+"""
+Health check for SR, to be triggered periodically by a systemd timer. What is checked is
+SR implementation type dependent.
+"""
+
+import SR
+import util
+
+
+def check_iscsi_sessions():
```
rosslagerwall, line 27: This feels a bit weird to me. The function is about checking iSCSI sessions but then it calls a generic check_sr() method. Would it be better to just call check_sr() on every SR and then let the SR decide what to check (if anything)?

rosslagerwall (approved):

MarkSymsCtx:
I've just realised that we already have `test_LVHDoISCSISR.py` so I'll move the tests in this into that.

rosslagerwall (approved):

drivers/sr_health_check.py
```diff
@@ -0,0 +1,56 @@
...
+        if not pbds:
+            continue
+
+        pbd_ref, pbd = pbds.popitem()
+        if not pbd['currently_attached']:
+            continue
+
+        sr_uuid = srs[sr]['uuid']
+        sr_obj = SR.SR.from_uuid(session, sr_uuid)
+        sr_obj.check_sr(sr_uuid)
```
edwintorok, line 52: What if you have a SMAPIv3 SR here, will that work, or do you need to filter them out?

edwintorok: You can find a list of SMAPIv1 SRs with something like this (can be translated into python API calls instead of xe):
`xe sm-list required-api-version=1.0 params=type --minimal|tr ',' '\n'|xargs -n1 -I% xe sr-list type=% --minimal|tr ',' '\n'|xargs echo`

MarkSymsCtx: That's a very good point, originally it only looked for `lvmoiscsi` SRs and then after addressing https://github.com/xapi-project/sm/pull/627#discussion_r1268376378 it exposes that problem.

TimSmithCtx (approved):

PR #631, opened 2023-08-18
Fix vdi-ref when static vdis are used

gthvn1:
When static vdis are used there is no snapshots and we don't want to call method from XAPI.

### DIFF
--- a/drivers/LVHDSR.py
+++ b/drivers/LVHDSR.py
@@ -1509,10 +1509,11 @@ class LVHDVDI(VDI.VDI):
         elif self.sr.provision == "thick":
             needDeflate = False
             # except for snapshots, which are always deflated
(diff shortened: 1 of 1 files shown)


gthvn1:
This fixes an issue I saw when I ran `quicktest` on LVMoISICI SR. When `quicktest` ran the *static-vdis* test it fails because it tries to get the *vdi-ref* from XAPI and not from the config file.

MarkSymsCtx:
I think this is probably OK, but I'll pull it into the internal build system and run some tests on it before approving it.

gthvn1:
Ok. Thanks a lot.

MarkSymsCtx:
> I think this is probably OK, but I'll pull it into the internal build system and run some tests on it before approving it.

Tests passed OK.

KevinLCtx (approved):

PR #648, opened 2023-10-17
Remove SR lock during thin attach/detach

benjamreis:
This lock is normally useless and can create a dead lock when thin mode is activated:
- A task try to deactivate a volume during a VM shutdown on a slave (so a VDI A is locked). Then a new task is created on the master host, we try to get the SR lock on the master.
- In parallel a tap-pause is asked from the master to the slave, the master SR lock is now locked. The tap-pause request is received on the slave, but we can't lock VDI A because it's already locked.

So to resume: a dead lock is only possible if we try to shutdown a VM with a particular VDI and if we try to snapshot it in the same time.

### DIFF
--- a/drivers/lvhdutil.py
+++ b/drivers/lvhdutil.py
@@ -213,25 +213,11 @@ def setSizeVirt(journaler, srUuid, vdiUuid, size, jFile):
     vhdutil.setSizeVirt(path, size, jFile)
 
 
-def _tryAcquire(lock):
-    """We must give up if the SR is locked because it could be locked by the
-    coalesce thread trying to acquire the VDI lock we're holding, so as to
-    avoid deadlock"""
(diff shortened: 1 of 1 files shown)


benjamreis:
It is an issue we found out on [our Linstor driver](https://github.com/xcp-ng/sm/commit/b2bef4a41ccd46d2f7f6266333d07ce40977dc8a) that also applies to this one.

Unless the lock is necessary to activate the LVM cache, in this case we can re add it only for the cache.

MarkSymsCtx:
> So to summarize: a dead lock is only possible if we try to shutdown a VM with a particular VDI and if we try to snapshot it in the same time.
This sounds like the solution would be for the initiating process on the coordinator to hold a lock on the VDI whilst performing these operations and not permit the detach and snapshot operations to occur in parallel, that sounds like an unending recipe for edge cases.

PR #651, opened 2023-10-18
Optimized sub-optimal code.

ProgramR4732:
Made the following changes:

-> replaced un-changed lists by tuples for better runtime & memory savings.
-> changed LVHDoHBASR.handles method argument type -> type_ as it coincides with the core type function. 
-> converted % formatted strings into f-string for better readability, and runtime.
-> removed unneccessary if-else block in LVHDoHBASR.handles method.
-> removed unneccessary line-break '\' references.
-> some other minor formatting fixes.

### DIFF
--- a/drivers/LVHDoHBASR.py
+++ b/drivers/LVHDoHBASR.py
@@ -34,14 +34,16 @@ import scsiutil
 import mpath_cli
 import glob
 
-CAPABILITIES = ["SR_PROBE", "SR_UPDATE", "SR_METADATA", "SR_TRIM",
+CAPABILITIES = ("SR_PROBE", "SR_UPDATE", "SR_METADATA", "SR_TRIM",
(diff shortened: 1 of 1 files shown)


drivers/LVHDoHBASR.py
```diff
@@ -51,31 +53,29 @@
...
         else:
             name = __name__
-        if name.endswith("LVMoHBASR"):
-            return type == "lvmohba"  # for the initial switch from LVM
-        if type == "lvhdohba":
-            return True
-        return False
+
+        # for the initial switch from LVM
+        return type_ == "lvmohba" if name.endswith("LVMoHBASR") else "lvhdohba"
```
edwintorok, line 70: This needs some parenthesis, as it is currently written it'll return either a boolean or a string.
Some 'mypy' type annotations might be useful in catching this kind of bug.

PR #655, opened 2023-11-02
CA-384783 Probe for NFS4 when rpcinfo does not include it

TimSmithCtx:
Just because the rpcinfo list obtained from rpcbind does not contain NFS4 does not mean it is not supported. Probe for it if it is not in the list. Some filers supporting both do not register NFS4 with rpcbind.

### DIFF
--- a/drivers/nfs.py
+++ b/drivers/nfs.py
@@ -81,10 +81,14 @@ def check_server_service(server, transport):
     NFS_SERVICE_RETRY * NFS_SERVICE_WAIT
     """
 
-    sv = get_supported_nfs_versions(server, transport)
-    # Services are not present in NFS4 only, this doesn't mean there's no NFS
-    if sv == ['4']:
-        return True
+    try:
+        sv = get_supported_nfs_versions(server, transport)
+        # Services are not present in NFS4 only, this doesn't mean there's no NFS
+        if "4" in sv:
+            return True
+    except NfsException:
+        # Server failed to give us supported versions
+        pass
 
     retries = 0
     errlist = [errno.EPERM, errno.EPIPE, errno.EIO]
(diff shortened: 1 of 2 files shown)


edwintorok (commented):
What happens with a server that supports NFSv3 only, will the NFS4 probe potentially take a long time? (do we have a timeout?)

TimSmithCtx:
> What happens with a server that supports NFSv3 only, will the NFS4 probe potentially take a long time to fail? (do we have a timeout?)

Since the probe is an attempted mount with a specific NFS version, in that case the NFS server should respond very quickly with "No. Go away."

MarkSymsCtx (approved):

MarkSymsCtx:
@benjamreis you should be aware of this, a bug from your recent changes. We have a customer who has a filer (TrueNAS) which supports NFSv4 but only returns 2 & 3 in the response from `rpcinfo`

olivierlambert:
Why TrueNAS (which version? Core or Scale?) would returns 2&3 from `rpcinfo` while supporting only v4? Have you made a report to TrueNAS?

MarkSymsCtx:
> Why TrueNAS (which version? Core or Scale?) would returns 2&3 from `rpcinfo` while supporting also v4? Have you made a report to TrueNAS?
> 
> edit: do you mean that if your TrueNAS (question on the version still stands) supports 2,3 and 4, it will reports only 2 and 3? Is it an expected behavior? Because if not, it's not a bug here then.

It is not required to return 4 in response to rpcinfo, some filers do, some do not even when they support NFS v4.

olivierlambert:
Okay that makes sense to also check it then. What's the functional issue with this, outside -I suppose- only mount it in v3 while v4 is also available? I'm not sure to understand the "bug" outside not using v4 while it's available but not announced?

alexbrett:
The functional issue was the user was specifying to use v4 in the sr-create, and were getting an exception back saying it was an unsupported version, when prior to the recent changes this had worked fine, so it was a regression from their point of view.

olivierlambert:
That makes sense, thanks!

rdn32 (approved):

KevinLCtx (approved):

PR #658, opened 2023-11-28
CP-39600: Rework LVM locking to use fair lock queue

TimSmithCtx:
This is intended to prevent a storm of later LVM commands from usurping the lock that an earlier command is waiting for, just because they happened to manage to start waiting before the original holder was done.

In some pathalogical circumstances, this has been known to result in a commant taking many minutes to get the lock.

### DIFF
--- a/drivers/LVHDSR.py
+++ b/drivers/LVHDSR.py
@@ -1758,9 +1758,7 @@ class LVHDVDI(VDI.VDI):
         if snapType == VDI.SNAPSHOT_DOUBLE:
             clonUuid = util.gen_uuid()
         jval = "%s_%s" % (baseUuid, clonUuid)
-        with lvutil.LvmLockContext():
-            # This makes multiple LVM calls so take the lock early
-            self.sr.journaler.create(self.JRN_CLONE, origUuid, jval)
+        self.sr.journaler.create(self.JRN_CLONE, origUuid, jval)
         util.fistpoint.activate("LVHDRT_clone_vdi_after_create_journal", self.sr.uuid)
 
         try:
@@ -1835,9 +1833,7 @@ class LVHDVDI(VDI.VDI):
             self._failClone(origUuid, jval, str(e))
         util.fistpoint.activate("LVHDRT_clone_vdi_before_remove_journal", self.sr.uuid)
 
-        with lvutil.LvmLockContext():
-            # This makes multiple LVM calls so take the lock early
-            self.sr.journaler.remove(self.JRN_CLONE, origUuid)
+        self.sr.journaler.remove(self.JRN_CLONE, origUuid)
 
         return self._finishSnapshot(snapVDI, snapVDI2, hostRefs, cloneOp, snapType)
 
--- a/drivers/lock.py
+++ b/drivers/lock.py
@@ -49,7 +49,6 @@ class Lock(object):
             instances[name] = LockImplementation(name, ns)
         return instances[name]
 
-    # These are required to pacify pylint as it doesn't understand the __new__
     def acquire(self):
         raise NotImplementedError("Lock methods implemented in LockImplementation")
 
--- a/drivers/lock_queue.py
+++ b/drivers/lock_queue.py
@@ -1,0 +1,120 @@
+# Copyright (C) Cloud Software Group, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation; version 2.1 only.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+import os
+import pickle
+import sys
+import time
+
+import lock
+import util
+
+
+DEBUG_LOG = True
+
+def debug_log(msg):
+    if DEBUG_LOG:
+        util.SMlog("LockQueue: " + msg)
+
+def get_process_start_time(pid):
+    proc_file = f"/proc/{pid}/stat"
+    with open(proc_file, 'r') as f:
+        return f.read().split(')')[-1].split(' ')[20]
+
+def process_is_valid(pid, start_time):
+    proc_file = f"/proc/{pid}/stat"
+
+    try:
+        if start_time != get_process_start_time(pid):
+            debug_log(f"Process {pid} has incorrect start time real:{get_process_start_time(pid)} vs expected:{start_time}")
+            return False
+    except FileNotFoundError:
+        debug_log(f"Process {pid} is dead")
+        return False
+
+    return True
+
+class LockQueue:
+    def __init__(self, name):
+        self.name = name
+        self._queue_lock = lock.Lock(name, f"ql-{name}")
+        self._action_lock = lock.Lock(name, f"al-{name}")
+        # Filename to hold the process queue
+        self._mem = f"/tmp/mem-{name}"
+
+    def load_queue(self):
+        try:
+            with open(self._mem, "rb") as f:
+                queue = pickle.load(f)
+            debug_log("load_queue {}".format(queue))
+        except EOFError:
+            queue = []
+        except FileNotFoundError:
+            queue = []
+        return queue
+
+    def save_queue(self, queue):
+        with open(self._mem, "w+b") as f:
+            pickle.dump(queue, f)
+        debug_log("save_queue {}".format(queue))
+
+    def push_into_process_queue(self):
+        self._queue_lock.acquire()
+
+        queue = self.load_queue()
+        queue.append((os.getpid(), get_process_start_time(os.getpid())))
+        self.save_queue(queue)
+
+        self._queue_lock.release()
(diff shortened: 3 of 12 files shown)


drivers/LVHDSR.py
```diff
@@ -1758,9 +1758,8 @@ def _snapshot(self, snapType, cloneOp=False, cbtlog=None, cbt_consistency=None):
         if snapType == VDI.SNAPSHOT_DOUBLE:
             clonUuid = util.gen_uuid()
         jval = "%s_%s" % (baseUuid, clonUuid)
-        with lvutil.LvmLockContext():
-            # This makes multiple LVM calls so take the lock early
-            self.sr.journaler.create(self.JRN_CLONE, origUuid, jval)
+        # This makes multiple LVM calls so take the lock early
```
MarkSymsCtx, line 1761: Redundant comment

drivers/LVHDSR.py
```diff
@@ -1835,9 +1834,8 @@ def _snapshot(self, snapType, cloneOp=False, cbtlog=None, cbt_consistency=None):
             self._failClone(origUuid, jval, str(e))
         util.fistpoint.activate("LVHDRT_clone_vdi_before_remove_journal", self.sr.uuid)
 
-        with lvutil.LvmLockContext():
-            # This makes multiple LVM calls so take the lock early
-            self.sr.journaler.remove(self.JRN_CLONE, origUuid)
+        # This makes multiple LVM calls so take the lock early
+        self.sr.journaler.remove(self.JRN_CLONE, origUuid)
```
MarkSymsCtx, line 1838: And another

drivers/lock.py
```diff
@@ -53,6 +53,7 @@ def __new__(cls, name, ns=None, *args, **kwargs):
     def acquire(self):
         raise NotImplementedError("Lock methods implemented in LockImplementation")
 
+    # These are required to pacify pylint as it doesn't understand the __new__
```
MarkSymsCtx, line 56: Probably don't need this change as it's only a comment?

drivers/lock_queue.py
```diff
@@ -0,0 +1,123 @@
...
+        self.push_into_process_queue()
+
+        # Keep reading the process queue until we are at the front
+        while True:
+            self._queue_lock.acquire()
+            queue = self.load_queue()
+            front_pid, front_start_time = queue.pop(0)
+            print(f"Testing for PID {front_pid}")
+            if front_pid == os.getpid():
+                # We are at the front, it is now our turn to do stuff
```
MarkSymsCtx, line 96: 'to do our work' maybe? And below in the actual action phase

drivers/lock_queue.py
```diff
@@ -0,0 +1,123 @@
...
+                self._queue_lock.release()
+                break
+
+            # Getting here means it was not our turn to do stuff
+            # If the process at the front of the queue is not alive then remove it
+            if not process_is_valid(front_pid, front_start_time):
+                # front pid has already been popped from queue so just save it
+                debug_log(f"Removing invalid process {front_pid}")
+                self.save_queue(queue)
+            # Release the lock and try again later. Most waiting will be on the queue lock,
```
MarkSymsCtx, line 111: Add a blank line above this

drivers/lvutil.py
```diff
@@ -238,7 +255,7 @@ def _checkVG(vgname):
     try:
         cmd_lvm([CMD_VGS, "--readonly", vgname])
         return True
-    except:
+    except Exception as e:
```
MarkSymsCtx, line 258: `e` is not being used

tests/test_LVHDSR.py
```diff
@@ -4,6 +4,7 @@
 
 import uuid
 
+import lock_queue
```
MarkSymsCtx, line 7: Is this actually necessary?

tests/test_lock_queue.py
```diff
@@ -0,0 +1,117 @@
...
+    @mock.patch('lock_queue.pickle.dump', side_effect=mock_pickle_dump_fn)
+    @mock.patch('lock_queue.os.getpid')
+    @mock.patch('lock_queue.get_process_start_time')
+    @mock.patch('lock.Lock', autospec=False)
+    def test_push_to_queue_3x(self, lock, start_time, getpid, pdump, pload):
+        global saved_queue
+
+        lq = lock_queue.LockQueue(self.get_lock_name())
+        # Push to queue 3 times using these PID and Start Time combinations
+        test_pids = [993, 996, 997]
```
MarkSymsCtx, line 67: might be better if the pids were not in increasing order here. I'm sure the code functions correctly but it would make it clear and guard against any future change which accidentally sorted the queue for instance.

MarkSymsCtx (changes requested):
Some minor changes

MarkSymsCtx (approved):

drivers/lock_queue.py
```diff
@@ -0,0 +1,121 @@
...
+
+class LockQueue:
+    def __init__(self, name):
+        self.name = name
+        self._queue_lock = lock.Lock(name, f"ql-{name}")
+        self._action_lock = lock.Lock(name, f"al-{name}")
+
+        # Open file to hold the process queue
+        self._mem = f"/tmp/mem-{name}"
+        if not os.path.exists(self._mem):
```
rdn32, line 59: Surely there is a race condition here. Two processes might simultaneously find that the queue file doesn't exist, and both try to initialize it. So one might set it to be empty, even this were after the other had enqueued itself.

drivers/lock_queue.py
```diff
@@ -0,0 +1,121 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# Miscellaneous utility functions
```
rdn32, line 16: Update comment or remove it.

tests/test_lock_queue.py
```diff
@@ -0,0 +1,117 @@
...
+import os
+import sys
+import unittest
+import unittest.mock as mock
+
+import lock
+import lock_queue
+
+## Instead of saving the process queue to disk the mocks will save it here.
+## It needs to be global because it is shared between threads.
```
rdn32, line 12: I didn't see any threading in the tests

tests/test_lock_queue.py
```diff
@@ -0,0 +1,117 @@
...
+class Test_LockQueue(unittest.TestCase):
+    def setUp(self):
+        # Re-initialize queue to empty for each test
+        global saved_queue
+        saved_queue = []
+
+    def get_lock_name(self):
+        return "bacon"
+
+    def tearDown(self):
```
rdn32, line 33: Since this doesn't do anything, I assume it can be removed

tests/test_lock_queue.py
```diff
@@ -0,0 +1,117 @@
...
+        start_time.return_value = 575
+
+        # Queue is empty
+        self.assertEqual(saved_queue, [])
+
+        with lock_queue.LockQueue(self.get_lock_name()) as lq:
+            # Should have removed from the queue before completing entry to the context manager
+            self.assertEqual(saved_queue, [])
+
+        # LockQueue removes itself from queue when exiting the context manager
```
rdn32, line 95: All the adding to and removing from the queue happens when entering the context manager, surely.

drivers/lock_queue.py
```diff
@@ -0,0 +1,121 @@
...
+        while True:
+            self._queue_lock.acquire()
+            queue = self.load_queue()
+            front_pid, front_start_time = queue.pop(0)
+            print(f"Testing for PID {front_pid}")
+            if front_pid == os.getpid():
+                # We are at the front, it is now our turn to wait on the action lock
+                # and then do our work
+                debug_log(f"{front_pid} taking action lock")
+                self._action_lock.acquire()
```
rdn32, line 97: If another process already holds the action lock, then the current process will have to wait until it can release the queue lock. In which case, any other processes wanting to take out locks won't even get onto the queue until that happens, and the order in which they end up on the queue will depend on flock, which I thought were didn't trust on that score. I don't know how this would pan out in practice, but the locking might not be as fair as we hope.

rdn32 (changes requested):
The only serious thing was that I think I saw a potential race condition in LockQueue's constructor, which looks like it could be readily fixed with the queue lock.

drivers/lock_queue.py
TimSmithCtx: Consider what happens when the process holding the action lock is killed. As little as possible should be required of other processes when that happens. In this case, nothing is required as there is nothing to clean up, and no handlers need to be installed to clean up after non-KILL signals. Cleanup only needs to worry about waiters who died while waiting, which should be much less common.

Also, your solution would require regular checking of the queue head while the action lock is held, which basically means you've got a load of processes constantly checking after short sleeps instead of properly asleep in the kernel. You've implemented a spinlock. Let's not do that.

I am well aware that this mechanism isn't perfectly fair. I proposed a method which is, and only requires one lock, but it's not what got implemented (arguably its check is too heavyweight). So it goes. In the case of a lot of queued waiters, processes already in the queue will sleep between checks where new arrivals won't. so that should help.

tests/test_lock_queue.py
rdn32: No, but "is" is present tense

rdn32 (dismissed):

MarkSymsCtx (approved):

PR #659, opened 2023-12-06
CP-46863 Dump Multipath Status from Storage Manager

LunfanZhang:
This PR is **not intended for merging** but proposes a prototype for dumping the Multipath status. The purpose is to discuss and reach a conclusion for a final design based on this prototype. Once there is no further discussion, I will submit another PR or convert it into a formal one. which will involve NVMe/FC as well.
Basically, for this scenario we can dump the data to a normal file and also to query it with IPC, considering that NRPE does not care about every multipath status change but only the latest status, so I presume IPC is not the suitable way to implement rather than dump the data to a normal file.
and another way like Shared Memory can also resolved this issue but it is not safe as the NRPE service is not privilege process.
Therefore, I've chosen the simplest and secure approach for now, dumping the multipath status to a normal file as below, 
since it`s my first time to submit PR to SM, every comment or feedback is welcome.

### DIFF
--- a/drivers/mpathcount.py
+++ b/drivers/mpathcount.py
@@ -21,6 +21,9 @@ import sys
 import re
 import xs_errors
 import mpath_cli
+import json
+import tempfile
+import shutil
 
 supported = ['iscsi', 'lvmoiscsi', 'rawhba', 'lvmohba', 'ocfsohba', 'ocfsoiscsi', 'netapp', 'lvmofcoe', 'gfs2']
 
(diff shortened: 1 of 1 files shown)


drivers/mpathcount.py
```diff
@@ -246,7 +265,8 @@ def add(key, val):
             if srtype in supported:
                 devconfig = record["device_config"]
                 sm_config = session.xenapi.SR.get_sm_config(SR)
-                check_devconfig(devconfig, sm_config, config, remove, add)
+                check_devconfig(devconfig, sm_config, config, remove, add, mpath_status)
+        safe_dump_iscsi_mp_status(mpath_status)
```
MarkSymsCtx, line 269: This should be indented to the same level as the check_devconfig above it

MarkSymsCtx (commented):
Looks reasonable, one minor thing about indentation.

PR #662, opened 2023-12-12
CP-46863 Dump Multipath Status from Storage Manager

LunfanZhang:
- Save the multi-path status to a regular file whenever a Udev event is triggered.
- Generate the 'mp_status' file if it does not exist when starting the 'mpathcount' service for the first time.

### DIFF
--- a/drivers/mpathcount.py
+++ b/drivers/mpathcount.py
@@ -21,6 +21,7 @@ import sys
 import re
 import xs_errors
 import mpath_cli
+import json
 
 supported = ['iscsi', 'lvmoiscsi', 'rawhba', 'lvmohba', 'ocfsohba', 'ocfsoiscsi', 'netapp', 'lvmofcoe', 'gfs2']
 
@@ -29,6 +30,8 @@ LOCK_NS1 = "mpathcount1"
 LOCK_NS2 = "mpathcount2"
 
 MAPPER_DIR = "/dev/mapper"
+MPATHS_DIR = "/dev/shm"
+MPATH_FILE_NAME = "/dev/shm/mpath_status"
 match_bySCSIid = False
 mpath_enabled = True
 SCSIid = 'NOTSUPPLIED'
@@ -113,7 +116,8 @@ def get_root_dev_major():
 # @entry:   string representing previous value
 # @remove:  callback to remove key
 # @add:     callback to add key/value pair
-def update_config(key, SCSIid, entry, remove, add):
+# @mpath_status:  map to record multipath status
+def update_config(key, SCSIid, entry, remove, add, mpath_status=None):
     path = os.path.join(MAPPER_DIR, SCSIid)
     util.SMlog("MPATH: Updating entry for [%s], current: %s" % (SCSIid, entry))
     if os.path.exists(path):
@@ -136,6 +140,8 @@ def update_config(key, SCSIid, entry, remove, add):
             add('multipathed', 'true')
             add(key, str(newentry))
             util.SMlog("MPATH: Set val: %s" % str(newentry))
+        if mpath_status != None:
+            mpath_status.update({str(key): f"{count}/{max}"})
     else:
         util.SMlog('MPATH: device %s gone' % (SCSIid))
         remove('multipathed')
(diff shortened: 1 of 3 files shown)


systemd/mpathcount.service
```diff
@@ -8,4 +8,5 @@ StandardInput=socket
 StandardOutput=null
 StandardError=journal
 ExecStart=/usr/bin/sh -c '. /etc/xensource-inventory; while dd of=/dev/null bs=4096 count=1 status=none conv=noerror; do /opt/xensource/sm/mpathcount.py; done'
+ExecStartPost=/bin/bash -c 'if [ ! -f "/dev/shm/mpath_status" ]; then /usr/bin/python3 /opt/xensource/sm/mpathcount.py; fi'
```
DeliZhangX, line 11: Maybe you can exec `/opt/xensource/sm/mpathcount.py` directly  when service restart no matter /dev/shm/mpath_status existing or not.

LunfanZhang: It will slightly slow down the startup of the mpathcount service.

DeliZhangX: What I concern is the mapth_status file won't update if the service stopped while udev notifies update event.
Leave to storage experts to help to review as well.

DeliZhangX (approved):

drivers/mpathcount.py
```diff
@@ -246,7 +253,9 @@ def add(key, val):
             if srtype in supported:
                 devconfig = record["device_config"]
                 sm_config = session.xenapi.SR.get_sm_config(SR)
-                check_devconfig(devconfig, sm_config, config, remove, add)
+                check_devconfig(devconfig, sm_config, config, remove, add, mpath_status)
+        util.atomicFileWrite(MPATH_FILE_NAME, MPATHS_DIR, json.dumps(mpath_status))
```
MarkSymsCtx, line 257: These two lines should be indented to match the check_devconfig above

systemd/mpathcount.service
MarkSymsCtx: You don't need this, certainly do not need to call it with `/usr/bin/python3` as the mpathcount.py has a `#!/usr/bin/python3` already

MarkSymsCtx (changes requested):

drivers/mpathcount.py
LunfanZhang: This file `MPATH_FILE_NAME` is created for record the multi-path status per host, and it dumps the status of all PBD's paths. therefore, I think these two lines may not be indented to match `check_devconfig ` ?

systemd/mpathcount.service
LunfanZhang: Agree! I have deleted it.

systemd/mpathcount.service
```diff
@@ -8,4 +8,5 @@ StandardInput=socket
 StandardOutput=null
 StandardError=journal
 ExecStart=/usr/bin/sh -c '. /etc/xensource-inventory; while dd of=/dev/null bs=4096 count=1 status=none conv=noerror; do /opt/xensource/sm/mpathcount.py; done'
+ExecStartPost=/bin/bash -c 'if [ ! -f "/dev/shm/mpath_status" ]; then /opt/xensource/sm/mpathcount.py; fi'
```
MarkSymsCtx, line 11: I think this, if we do indeed need it, should `ExecStartPre` so that it will not interfere with the pipeline in the `ExecStart`

MarkSymsCtx (changes requested):
One minor change I think

systemd/mpathcount.service
LunfanZhang: I believe we need it. After RPU, if there is no path change event, NRPE or some component like NRPE won't detect the path status. Yes, I can move it to ExecStartPre. That's okay with me.

LunfanZhang: Hi @MarkSymsCtx I have fixed this issue. could we continue to move on?

MarkSymsCtx (approved):

KevinLCtx (approved):

PR #671, opened 2024-02-05
CA-385069: Remove unnecessary LvmContext wrap

MarkSymsCtx:
This was resulting in a deadlock condition being detected and is no longer required after the chnages for CP-39600 in 125aa9e.


MarkSymsCtx:
Mostly it's whitespace after removing the `with` wrap from the code block

TimSmithCtx (approved):

PR #672, opened 2024-02-05
CA-388451: ensure that xapi sessions are logged out

MarkSymsCtx:


lindig (approved):
Is there a `with_session` function? Python seems to go in that direction. But this should work.

MarkSymsCtx:
> Is there a `with_session` function? Python seems to go in that direction. But this should work.

A context_manager would be the pythonic way of doing it but I don't think XenAPI.py provides one.

robhoes (approved):

rdn32 (approved):

TimSmithCtx (approved):

PR #673, opened 2024-02-05
Always remove the RO/RW tag from VDIs in case of failure

Wescoeur:
During VDI activation in the blktap module and in case of failure in "sm.VDI.from_uuid" call, the RO/RW tag is never removed. As a result a VDI can no longer be used correctly: an assert is triggered each time we try to re-activate this volume because the tag is still present.

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -1602,20 +1602,24 @@ class VDI(object):
         """Wraps target.activate and adds a tapdisk"""
 
         #util.SMlog("VDI.activate %s" % vdi_uuid)
+        refresh = False
         if self.tap_wanted():
             if not self._add_tag(vdi_uuid, not options["rdonly"]):
                 return False
-            # it is possible that while the VDI was paused some of its
-            # attributes have changed (e.g. its size if it was inflated; or its
-            # path if it was leaf-coalesced onto a raw LV), so refresh the
-            # object completely
-            params = self.target.vdi.sr.srcmd.params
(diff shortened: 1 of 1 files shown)


MarkSymsCtx (approved):

drivers/blktap2.py
```diff
@@ -1615,7 +1617,6 @@ def _activate_locked(self, sr_uuid, vdi_uuid, options):
             driver_info = target.sr.srcmd.driver_info
```
rdn32, line 1617: These few lines of code used to be conditional on `self.tap_wanted()` but now they're run unconditionally. Is that deliberate? Or an accidental side-effect of the way you've moved the beginning of the `try` block?

MarkSymsCtx: well spotted, that's a bug

MarkSymsCtx (changes requested):

drivers/blktap2.py
Wescoeur: I pushed this change a little too quickly... So fixed I guess this time. :)

rdn32 (approved):

MarkSymsCtx:
Being picky it's not actually the `pause tag` you're ensuring is removed by rather the active on host RW/RO key. The change itself looks fine and I'm not overly bothered if the commit message is a bit misleading.

MarkSymsCtx (approved):

Wescoeur:
Right. I modified the description.

TimSmithCtx (approved):

PR #674, opened 2024-02-07
Add unittests to exercise Xapi session handling

MarkSymsCtx:


tests/test_util.py
```diff
@@ -713,3 +713,62 @@ def test_unictrunc(self):
...
+        self.assertIn('Unknown fist point: TestInvalidFP',
+                      self.mock_log.call_args[0][0])
+
+    def test_activate_not_active(self):
+        test_uuid = str(uuid.uuid4())
+        valid_fp_name = "TestValidFP"
+
+        fistpoints = util.FistPoint([valid_fp_name])
+
+        fistpoints.activate(valid_fp_name, test_uuid)
```
KevinLCtx, line 758: It's not obvious what this is testing for.

MarkSymsCtx: Fair, if the FIST isn't active then nothing should be actioned. I'll add negative versions of the asserts in `test_activate_not_exit` to make that clear

KevinLCtx (approved):

PR #676, opened 2024-02-15
CP-47841: update multipath configuration for PURE Storage

MarkSymsCtx:


KevinLCtx (approved):

multipath/multipath.conf
```diff
@@ -142,4 +142,17 @@ devices {
         failback                    immediate
         no_path_retry               18
     }
+    device {
```
rosslagerwall, line 145: Where did this configuration come from?

It looks similar to but not the same as the upstream built in configuration:
```
        {
                /* FlashArray */
                .vendor        = "PURE",
                .product       = "FlashArray",
                .pgpolicy      = GROUP_BY_PRIO,
                .pgfailback    = -FAILBACK_IMMEDIATE,
                .hwhandler     = "1 alua",
                .prio_name     = PRIO_ALUA,
                .fast_io_fail  = 10,
                .max_sectors_kb = 4096,
        },
```

In particular, it is missing the max_sectors_kb which is explicitly called out in the commit message:

```
commit 95030d2c11a12c186e5ded3239dc05973f9359c8
Author: Brian Bunker <brian@purestorage.com>
Date:   Mon Mar 2 15:53:54 2020 +0100

    ALUA support for PURE FlashArray

    Hello all,

    It has been some time since we updated our PURE FlashArray
    configuration. The Linux vendors that we had been seeing in the field
    were using very old versions of multipath-tools, so we haven’t needed to
    change anything for some time. With the release of RHEL8, some of our
    earlier values have been lost by upstream changes.

    In addition we have our Active Cluster feature which leverages ALUA
    since our last patch. The ALUA confguration will work for all
    FlashArrays with or without Active Cluster.

    We are changing 3 things.

    1. ALUA support
    2. Fast fail timeout from the default of 5 seconds to 10 seconds (We
       need this for our FC NPIV port migration).
    3. Maximum sector size of 4MB. Some Linux vendors don’t honor the block
       limits VPD page of INQUIRY).
...
```

(unless it is relying on defaults set somewhere?)

MarkSymsCtx: From the Vendor's documentation - https://support.purestorage.com/Solutions/Citrix/Citrix_XenServer/Pure_Storage_All-Flash_Storage_Array_and_Citrix_XenServer

rosslagerwall (approved):

TimSmithCtx (approved):

PR #683, opened 2024-04-18
CA-387861 Introduce fair locking subsystem

TimSmithCtx:
None of the file-based locking systems available are "fair", as in, there is no guarantee that the first waiter will not be usurped when the lock is released by later waiters, thus suffering from lock starvation. This has been something of a problem across various attempts to wrap access to LVM, and indeed it turns out that multipath commands also need to be serialised on some of the same resources.

This adds a small templated systemd service which does nothing but accept one connection at a time on a UNIX Domain socket, with a fairly long accept queue. The service is started when needed (thus every process using this mechanism needs permission to start the service), and the "Lock" is acquired by connecting to the socket. connect() on UNIX Domain sockets does not time out, and waiting connections are held in a queue, providing fairness.

The python3 context manager is provided via a metaclass which checks the arguments and allows only one of each named Fairlock object to exist, so that attempts to take a lock when it is already held can raise an exception. This is not thread-safe but could probably be made so.

### DIFF
--- a/misc/fairlock/fairlock.c
+++ b/misc/fairlock/fairlock.c
@@ -1,0 +1,57 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <errno.h>
+
+int main(int argc, char *argv[]) {
+     struct sockaddr_un addr;
+     int                sock;
+     int                fd;
+
+    if (argc < 2) {
+        fprintf(stderr, "Syntax: %s <socket filename>\n", argv[0]);
+        exit(1);
+    }
+
+    /* Unlink the socket just in case */
+    unlink(argv[1]);
+    /* Create and bind a unix-domain socket with the passed-in name, and a listen
+     * queue depth of 64 */
+    sock = socket(AF_UNIX, SOCK_STREAM, 0);
+    memset(&addr, 0, sizeof(struct sockaddr_un));
+    addr.sun_family = AF_UNIX;
+    strncpy(addr.sun_path, argv[1], sizeof(addr.sun_path) - 1);
+    if (bind(sock, (const struct sockaddr *) &addr, sizeof(struct sockaddr_un)) < 0) {
+        fprintf(stderr, "bind() failed on socket %s: %s", argv[1], strerror(errno));
+        exit(1);
+    }
+    if (listen(sock, 64) < 0) {
+        fprintf(stderr, "listen(64) failed on socket %s: %s", argv[1], strerror(errno));
+        exit(1);
+    }
+
+    /* Now we have a socket, enter an endless loop of:
+     * 1) Accept a connection
+     * 2) Do a blocking read on that connection until EOF or error
+     *    (each of which means the client went away)
+     * 3) Close the socket on which we accepted the connection and
+     *    accept another one.
+     * 
+     * Having a connection to this socket thus provides an exclusive condition
+     * for which the queueing is fully fair up to a queue depth of 64 waiters.
+     * With more than 64 waiters, new entrants to the queue may get ECONNREFUSED
+     * (as if the server isn't running) and need to sleep and retry.
+     * Closing the client connection will cause the read() to return 0, terminating
+     * the connection
+     */
+    while (1) {
+        while ((fd = accept(sock, NULL, NULL)) > -1) {
+            char buffer[128];
+
+            do {} while (read(fd, buffer, sizeof(buffer)) > 0);
+            close(fd);
+        }
+    }
+}
--- a/misc/fairlock/fairlock.py
+++ b/misc/fairlock/fairlock.py
@@ -1,0 +1,67 @@
+import os
+import socket
+import inspect
+import time
+
+SOCKDIR = "/run/fairlock"
+START_SERVICE_TIMEOUT_SECS = 2
+
+class SingletonWithArgs(type):
+    _instances = {}
+    _init = {}
+
+    def __init__(cls, name, bases, dct):
+        cls._init[cls] = dct.get('__init__', None)
+
+    def __call__(cls, *args, **kwargs):
+        init = cls._init[cls]
+        if init is not None:
+            key = (cls, frozenset(
+                    inspect.getcallargs(init, None, *args, **kwargs).items()))
+        else:
+            key = cls
+
+        if key not in cls._instances:
+            cls._instances[key] = super(SingletonWithArgs, cls).__call__(*args, **kwargs)
+        return cls._instances[key]
+
+class FairlockDeadlock(Exception):
+    pass
+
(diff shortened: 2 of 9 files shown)


MarkSymsCtx:
Probably should have an update to https://github.com/xapi-project/sm/blob/master/mk/sm.spec.in ?

MarkSymsCtx:
Needs an update to .github/workflow/main.yml to repeat the python path changes

edwintorok:
Were you mixing byte range and whole file locks (or attempt overlapping byte range locks) previously? That one is known to have fairness issues: https://rusty.ozlabs.org/2010/08/13/fcntl-lock-starvation-and-tdb.html
Although there is some evidence that flock is not fair https://code.jjb.cc/linux-flock-does-not-provide-fair-locking (pointing to potentially a bug in the kernel? 

In python there might be a simpler way to get a process-wide fair lock though by using shared pthread mutexes (this is simple if you want to share lock with child processes and more complicated if you want to share with other processes). https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock

Perhaps a Manager object could be used for sharing Lock objects? https://docs.python.org/3/library/multiprocessing.html#using-a-remote-manager

TimSmithCtx:
> Probably should have an update to https://github.com/xapi-project/sm/blob/master/mk/sm.spec.in ?

Added. Also fixed a couple of rpmlint issues (Should not specify BuildRoot, or reference RPM_BUILD_ROOT or %buildroot in the %build section)

TimSmithCtx:
> Were you mixing byte range and whole file locks (or attempt overlapping byte range locks) previously? That one is known to have fairness issues: https://rusty.ozlabs.org/2010/08/13/fcntl-lock-starvation-and-tdb.html Although there is some evidence that flock is not fair https://code.jjb.cc/linux-flock-does-not-provide-fair-locking (pointing to potentially a bug in the kernel?

Neither fcntl() nor flock() is fair. We've run into this issue more than once.

I wouldn't call it a bug unless you can find somewhere any sort of fairness promise which is being violated. I know of no such promise. We might *wish* they were fair, but wishes are sadly not ponies...

> In python there might be a simpler way to get a process-wide fair lock though by using shared pthread mutexes (this is simple if you want to share lock with child processes and more complicated if you want to share with other processes). https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock

In a single-threaded process, there is no need for locks at all. Since this code is single-threaded, the locking requirement is clearly between processes, so this won't help :-)

MarkSymsCtx (approved):

edwintorok:
> > In python there might be a simpler way to get a process-wide fair lock though by using shared pthread mutexes (this is simple if you want to share lock with child processes and more complicated if you want to share with other processes). https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Lock
> 
> In a single-threaded process, there is no need for locks at all. Since this code is single-threaded, the locking requirement is clearly between processes, so this won't help :-)

If you look at the implementation it is a bit smarter than that, it uses named semaphores (in /dev/shm) https://github.com/python/cpython/blob/main/Modules/_multiprocessing/semaphore.c, so it does work between processes.

TimSmithCtx:
> If you look at the implementation it is a bit smarter than that, it uses named semaphores (in /dev/shm) https://github.com/python/cpython/blob/main/Modules/_multiprocessing/semaphore.c, so it does work between processes.

POSIX states that "the highest priority thread that has been waiting the longest shall be unblocked" only when either SCHED_FIFO or SCHED_RR applies to the blocked process. Apart from that, I see no promise of fairness in POSIX or SYSV semaphores.

And we've already been round the loop trying to manage a queue with locks. This is simpler and cleaner for our current purposes.

misc/fairlock/fairlock.py
```diff
@@ -0,0 +1,67 @@
...
+                raise FairlockServiceTimeout(f"Timed out starting service {service}")
+
+    def __enter__(self):
+        if self.connected:
+            raise FairlockDeadlock(f"Deadlock on Fairlock resource '{self.name}'")
+
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        try:
+            self.sock.connect(self.sockname)
+        except (FileNotFoundError, ConnectionRefusedError):
```
rdn32, line 57: There's a comment in fairlock.c saying that we'd also get this error if there were more than 64 waiters. I assume that trying to start the service when it is already running would be harmless, but we are only retrying the connection once. I take it that in reality we wouldn't expect to have that many waiters, so there's no point doing anything more involved. Yes?

tests/test_fairlock.py
```diff
@@ -0,0 +1,96 @@
...
+        self.mock_os.system.side_effect = [0, 1, 0, 0, 1, 0]
+        self.mock_time.time.side_effect = [0, 0, 0, 0, 0, 0]
+
+        with self.assertRaises(FairlockDeadlock) as err:
+            with Fairlock("test") as l:
+                n = Fairlock("test")
+                self.assertEquals(l, n)
+                # Real code would use another 'with Fairlock("test")' here but we cannot
+                # do that because it insists on having a code block as a body, which would
+                # then not be reached, causing a "Test code not fully covered" failure
```
rdn32, line 95: Much of a muchness, but you can use `# pragma: no cover`

rdn32 (approved):

misc/fairlock/fairlock.c
```diff
@@ -0,0 +1,57 @@
...
+     * With more than 64 waiters, new entrants to the queue may get ECONNREFUSED
+     * (as if the server isn't running) and need to sleep and retry.
+     * Closing the client connection will cause the read() to return 0, terminating
+     * the connection
+     */
+    while (1) {
+        while ((fd = accept(sock, NULL, NULL)) > -1) {
+            char buffer[128];
+
+            do {} while (read(fd, buffer, sizeof(buffer)) > 0);
```
rosslagerwall, line 53: If -1 is returned and `errno == EINTR` for some reason, I expect that it should continue the loop rather than exiting otherwise the lock might be dropped too early.

rosslagerwall (approved):

misc/fairlock/fairlock.py
TimSmithCtx: Local experimentation suggests that python's connect() handles the case of >64 waiters on the client side quite comfortably. While connecting to a Unix domain socket when the listen queue is full is allowed to return ECONNREFUSED, in practice on linux is is EAGAIN, which python handles for us.

The check for ECONNREFUSED here is mostly to cover the one-time-only race condition where we are the second client in, and the service has been started by the first client in but has not quite got around to listening yet.

misc/fairlock/fairlock.c
TimSmithCtx: This is indeed important if one is going to handle any signals. Fortunately the only signals coming in here will be ones which terminate the process entirely :-)

PR #684, opened 2024-05-02
CA-384942: path to the VM CD drive has been resolved to real device

MarkSymsCtx:
The error handling for empty CD drive assumes a path containing '/dev/xapi/cd/' but the path has been resolved to the actual device before the open operation is attempted. Check for paths of the form '^/dev/sr' as well so that the correct error message is emitted.

### DIFF
--- a/drivers/blktap2.py
+++ b/drivers/blktap2.py
@@ -784,8 +784,8 @@ class Tapdisk(object):
         arg = cls.Arg.parse(arg)
         return cls.launch(arg.path, arg.type, False)
 
-    @classmethod
-    def cgclassify(cls, pid):
+    @staticmethod
+    def cgclassify(pid):
 
         # We dont provide any <controllers>:<path>
         # so cgclassify uses /etc/cgrules.conf which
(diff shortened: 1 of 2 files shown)


KevinLCtx (dismissed):

rdn32 (dismissed):

TimSmithCtx (dismissed):

drivers/blktap2.py
```diff
@@ -840,9 +835,8 @@ def launch_on_tap(cls, blktap, path, _type, options):
 
         except TapCtl.CommandFailure as ctl:
             util.logException(ctl)
-            if ('/dev/xapi/cd/' in path and
-                    'status' in ctl.info and
-                    ctl.info['status'] == 123):  # ENOMEDIUM (No medium found)
+            if (('/dev/xapi/cd/' in path or path.startswith('/dev/sr')) and
```
andyhhp, line 838: I know it was the existing behaviour, but surely that should be `path.startswith('/dev/xapi/cd')` too ?

KevinLCtx (approved):

andyhhp (approved):

TimSmithCtx (approved):

PR #685, opened 2024-05-07
Make use of new fair locking subsystem

TimSmithCtx:
Wrap calls to commands which need to access the device mapper in a context manager for a resource called "device-mapper" so that they get serialised in a fair manner (no queue-jumping like a lockfile would have).


MarkSymsCtx (dismissed):

KevinLCtx (dismissed):

TimSmithCtx:
Using a "-" in the resource name causes the socket name to contain a "/" after being passed through systemd's substitutions, which causes All Sorts Of Trouble, so don't do that...

MarkSymsCtx (approved):

PR #689, opened 2024-05-28
feat(LVHDSR): add a way to modify config of LVMs

Wescoeur:
With this change the driver supports a "lvm-conf" param on "other-config". For now The configuration is only used by "remove" calls from LVMCache.

Example to issue discards after a lvremove command:

> xe sr-param-set uuid=<SR_UUID> other-config:lvm-conf=issue_discards=1

And to remove the param:

> xe sr-param-remove uuid=<SR_UUID> param-name=other-config param-key=lvm-conf


Wescoeur:
I closed it. Integrated in XCP-ng, will be added in this repo in a few months after QCOW2 branch integration.

PR #691, opened 2024-06-06
CP-49720 Move LOCK_TYPE_RUNNING from cleanup.py to lock.py

TimSmithCtx:
In the process, rename LOCK_TYPE_RUNNING to LOCK_TYPE_GC_RUNNING to describe *what* is running when it's running. This matches LOCK_TYPE_GC_ACTIVE which remains in cleanup.py because it is only used there.

Rename the global variables in cleanup.py to describe *what* is Active/Running.

### DIFF
--- a/drivers/FileSR.py
+++ b/drivers/FileSR.py
@@ -31,7 +31,7 @@ import blktap2
 import time
 import glob
 from uuid import uuid4
-from lock import Lock, LOCK_TYPE_GC_RUNNING
+from lock import Lock
 import xmlrpc.client
 import XenAPI # pylint: disable=import-error
 from constants import CBTLOG_TAG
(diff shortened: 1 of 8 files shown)


drivers/iscsilib.py
```diff
@@ -54,7 +53,7 @@ def doexec_locked(cmd):
     """Executes via util.doexec the command specified whilst holding lock"""
     _lock = None
     if os.path.basename(cmd[0]) == 'iscsiadm':
-        _lock = lock.Lock(LOCK_TYPE_RUNNING, 'iscsiadm')
+        _lock = lock.Lock(lock.LOCK_TYPE_GC_RUNNING, 'iscsiadm')
```
MarkSymsCtx, line 56: This is now incorrect. IN this case the lock should be a unique `LOCK_TYPE_ISCSIADM_RUNNING` as it is a lock around usage of iscsiadm and nothing to do with the GC.

MarkSymsCtx (changes requested):

drivers/FileSR.py
```diff
@@ -31,7 +31,7 @@
 import time
 import glob
 from uuid import uuid4
-from lock import Lock
+from lock import Lock, LOCK_TYPE_GC_RUNNING
```
MarkSymsCtx, line 34: Still needed? I suspect not.

drivers/LVHDSR.py
```diff
@@ -36,7 +36,8 @@
 import cleanup
 import blktap2
 from journaler import Journaler
-from lock import Lock
+from lock import Lock, LOCK_TYPE_GC_RUNNING
+import cleanup
```
MarkSymsCtx, line 40: Surely it must already have had cleanup imported as it was calling `cleanup.start_gc`? And the `LOCK_TYPE_GC_RUNNING` is no longer required

drivers/lock.py
```diff
@@ -23,6 +23,8 @@
 
 VERBOSE = True
 
+LOCK_TYPE_GC_RUNNING = "gc_running"
```
MarkSymsCtx, line 26: Do we risk having a clash with a pre-existing runing process during an upgrade where the running GC process already has the `running` lock and then a new process tries, and succeeds, in acquiring the `gc_running` lock and assumes it has free reign over the GC process? That probably results in data corruption unless we ensure the upgrade aborts any in progress GC operations, which it does not by default do.

drivers/LVHDSR.py
MarkSymsCtx: In fact, yes, on line 36

drivers/lock.py
```diff
@@ -23,6 +23,9 @@
 
 VERBOSE = True
 
+# Still just called "running" for backwards compatibility
+LOCK_TYPE_GC_RUNNING = "running"
+LOCK_TYPE_ISCSIADM_RUNNING = "isciadm_running"
```
MarkSymsCtx, line 28: This is fine but we will get a slightly odd lock name of `iscsiadm.iscsiadm_running` as a result.

MarkSymsCtx (approved):

drivers/lock.py
TimSmithCtx: I'm fine with that. If in the future we encounter a lock named `newthing.iscsiadm_running` we'll know where to look for the cut & paste error...

KevinLCtx (approved):

PR #692, opened 2024-06-13
CP-49775 convert SMGC to systemd service

TimSmithCtx:
Add a templated oneshot service and use systemctl to start it. The systemctl command is invoked with "--no-wait" so that it returns immediately because the "oneshot" nature of the service would otherwise wait until the command completes.

Note that because the service is a oneshot with RemainAfterExit=no (so we can multiply start it), it will go from "activating" to "dead" without ever appearing to be "running".

### DIFF
--- a/drivers/LVHDSR.py
+++ b/drivers/LVHDSR.py
@@ -1290,7 +1290,7 @@ class LVHDSR(SR.SR):
 
     def _kickGC(self):
         util.SMlog("Kicking GC")
-        cleanup.start_gc(self.session, self.uuid)
+        cleanup.start_gc_service(self.uuid)
 
     def ensureCBTSpace(self):
         # Ensure we have space for at least one LV
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -3206,6 +3206,10 @@ def gc(session, srUuid, inBackground, dryRun=False):
 
 
 def start_gc(session, sr_uuid):
+    """
+    This function is used to try to start a backgrounded GC session by forking
+    the current process. If using the systemd version, call start_gc_service() instead.
+    """
     # don't bother if an instance already running (this is just an
     # optimization to reduce the overhead of forking a new process if we
     # don't have to, but the process will check the lock anyways)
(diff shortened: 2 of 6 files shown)


systemd/SMGC@.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Garbage Collector for SR %I
+DefaultDependencies=no
+RequiresMountsFor=/run/sr-mount/%I
```
MarkSymsCtx, line 4: Not true as this will run on LVM SRs as well which do not have a mount point. It needs to work without a `RequiresMountsFor` settng

MarkSymsCtx (changes requested):

systemd/SMGC@.service
TimSmithCtx: Ah yes. Cut&paste error. Had no actual effect on a direct "start" command which would explain why the tests all worked.

MarkSymsCtx (changes requested):

drivers/LVHDSR.py
```diff
@@ -892,7 +892,7 @@ def _handleInterruptedCoalesceLeaf(self):
         entries = self.journaler.getAll(cleanup.VDI.JRN_LEAF)
         if len(entries) > 0:
             util.SMlog("*** INTERRUPTED COALESCE-LEAF OP DETECTED ***")
-            cleanup.gc_force(self.session, self.uuid)
+            cleanup.start_gc_service(self.uuid)
```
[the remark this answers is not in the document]
TimSmithCtx: Would it be appropriate to keep using start_gc_service() but add a flag to leave out the `--no-wait` argument ?
That would make the `systemctl` not return until the "startup" procedure was over (i.e. until the command finishes, given the sort of command it is).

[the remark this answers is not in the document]
MarkSymsCtx: I don't think so as gc_force doesn't actually perform a coalesce, it "just" fixes up journals and clears garbage

drivers/LVHDSR.py
```diff
@@ -892,7 +892,7 @@ def _handleInterruptedCoalesceLeaf(self):
         entries = self.journaler.getAll(cleanup.VDI.JRN_LEAF)
         if len(entries) > 0:
             util.SMlog("*** INTERRUPTED COALESCE-LEAF OP DETECTED ***")
-            cleanup.gc_force(self.session, self.uuid)
+            cleanup.start_gc_service(self.uuid, True)
```
MarkSymsCtx, line 895: No, this really does not mean the same thing as `gc_force` just leave it as `gc_force`

MarkSymsCtx (changes requested):

drivers/cleanup.py
```diff
@@ -3263,15 +3285,17 @@ def gc_force(session, srUuid, force=False, dryRun=False, lockSR=False):
...
-        lockGCActive.release()
-        return False
-    return True
+    sr_uuid_esc = srUuid.replace("-", "\\x2d")
+    cmd=[ "/usr/bin/systemctl", "is-active", f"SMGC@{sr_uuid_esc}"]
+    result = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
+    state = result.stdout.decode('utf-8').rstrip()
+    if state == "activating" or state == "running":
+        return True
+    return False
```
MarkSymsCtx, line 3298: Looks good, do we need to consider the case where `systemd` is not being used? I'm inclined to say not as I think all consumers of this code are now using `systemd`

MarkSymsCtx (approved):

MarkSymsCtx (approved):

drivers/cleanup.py
```diff
@@ -3206,6 +3206,10 @@ def gc(session, srUuid, inBackground, dryRun=False):
 
 
 def start_gc(session, sr_uuid):
```
rdn32, line 3208: Why is this code being retained? I don't see any callers, and the implementation of get_state implies that we only expect the systemd version to be called?

TimSmithCtx: Mainly to make it easy for someone who wants to revert to a non-systemd method of doing things to do so
(yes, they'd have to reimplement get-state but the original method of doing so has lock-ordering issues, so leaving it would not have been a service)

rdn32 (approved):

PR #694, opened 2024-07-01
CA-393194: Fix pvremove failure

stephenchengCloud:
In lvmohba test, the test failed at
`SR.destroy(sr_ref) -> removeVG -> pvremove ` with the lvm error message "No devices to process".
The root cause is that the PV which was to be removed
was not really used as PV when it comes to multipathed devices. 
The LVM official document says:
"Since each LUN has multiple device nodes in /dev  that point to the same underlying data,
they all contain the same LVM metadata
and thus LVM commands will find the same metadata
multiple times and report them as duplicates.
These duplicate messages are only warnings
and do not mean the LVM operation has failed.
Rather, they are alerting the user that only one of the devices has been used as a physical volume and the others are being ignored."  
Please refer to https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/logical_volume_manager_administration/duplicate_pv_multipath#duplicate_pv_multipath

This fix is to find the real PV in a VG before removing the VG.

Test passed: 4038532

### DIFF
--- a/drivers/lvutil.py
+++ b/drivers/lvutil.py
@@ -529,6 +529,27 @@ def createVG(root, vgname):
 
     # End block
 
+def getPVsInVG(vgname):
+    # Get PVs in a specific VG
+    pvs_ret = cmd_lvm([CMD_PVS, '--separator', ' ', '--noheadings', '-o', 'pv_name,vg_name'])
+    
+    # Parse each line to extract PV and VG information
+    # No need to handle exceptions here, return empty list if any error
+    pvs_in_vg = []
+    lines = pvs_ret.strip().split('\n')
+    for line in lines:
+        # To avoid invalid return format
+        parts = line.split()
+        if len(parts) != 2:
+            util.SMlog("Warning: Invalid or empty line in pvs output: %s" % line)
+            continue
+        pv, vg = parts
+        if vg == vgname:
+            pvs_in_vg.append(pv)
+    
+    util.SMlog("PVs in VG %s: %s" % (vgname, pvs_in_vg))
+    return pvs_in_vg
+
 def removeVG(root, vgname):
     # Check PVs match VG
     try:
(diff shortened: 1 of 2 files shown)


drivers/lvutil.py
```diff
@@ -529,6 +529,22 @@ def createVG(root, vgname):
...
+def getPVsInVG(vgname):
+    # Get PVs in a specific VG
+    pvs_ret = cmd_lvm([CMD_PVS, '--separator', ' ', '--noheadings', '-o', 'pv_name,vg_name'])
+    
+    # Parse each line to extract PV and VG information
+    # No need to handle exceptions here, return empty list if any error
+    pvs_in_vg = []
+    lines = pvs_ret.strip().split('\n')
+    for line in lines:
+        pv, vg = line.split()
```
LunfanZhang, line 541: Can we guarantee that the output at _line:534_ consistently matches the pattern `/dev/pv_xxx vg_xxx ` in all scenarios? If not, here might be exceptions.

stephenchengCloud: The '-o' option in the command at line 534 specifies the fields to display.

LunfanZhang: I random selected a host and tested , and the output was: `/dev/sda3--noheadingsVG_XenStorage-6d5630c0-a016-b2b8-d5fa-e8faed85ca3a`

stephenchengCloud: There is a space ' ' before --noheadings.

MarkSymsCtx:
@rdn32 we probably want to backport this to 8.2 as I suspect it will resolve your partner issue.

MarkSymsCtx (approved):

TimSmithCtx (dismissed):

MarkSymsCtx:
Whilst this change is in itself fine we will need to add unit tests for the code modified in order to be able to release it. Either in this PR or in a follow-up.

stephenchengCloud:
> Whilst this change is in itself fine we will need to add unit tests for the code modified in order to be able to release it. Either in this PR or in a follow-up.
> 
> test_lvutil.py has separate classes for the various operations and currently does not have anything for VG/PV management

Sure. I will add unit tests in this PR.

stephenchengCloud:
Added unit tests. 
Passed:
drivers/lvutil.py                     511    260    134      9    47%

MarkSymsCtx (approved):

TimSmithCtx (approved):

PR #698, opened 2024-07-23
CA-395554 Improve fairlocking reliability

TimSmithCtx:
Under some circumstances (when the listen queue is short enough) a process's blocking connect() to a UNIX-domain stream socket can return success even when the server side has not yet called accept().

To work around this, we have the server portion write a small fixed blob of data to each new connection and require the client to read it before it considers itself to have the lock; merely being connected successfully is no longer sufficient.

### DIFF
--- a/misc/fairlock/fairlock.c
+++ b/misc/fairlock/fairlock.c
@@ -5,6 +5,7 @@
 #include <sys/un.h>
 #include <errno.h>
 #include <syslog.h>
+#include <signal.h>
 
 int main(int argc, char *argv[]) {
      struct sockaddr_un addr;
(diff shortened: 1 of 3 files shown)


MarkSymsCtx (dismissed):

KevinLCtx (approved):

MarkSymsCtx (approved):

rosslagerwall:
> Under some circumstances (when the listen queue is short enough) a process's blocking connect() to a UNIX-domain stream socket can return success even when the server side has not yet called accept().

FWIW, connect() on a UDS does not synchronize with accept(). The only time connect() would block is when the listen queue is full. In that case, it will block until the server calls accept() (and pulls a different, already connected, socket from the listen queue).

TimSmithCtx:
Yeah. It would be great if some knowledgeable individual would someday update the unix(7) manual page with this information, particularly since it violates the law of least astonishment. I can't find the behaviour mentioned anywhere.

PR #711, opened 2024-10-08
CA-400106: disable leaf coalesce with VDI snapshot secondary

MarkSymsCtx:
When the secondary/mirror option is provided to the vdi_snasphot operation we do not want any leaf coalesce garbage collection taking place. Unfortunately xapi will delete the snapshot once it's finished copying the new parent to the destination SR in SXM and whilst the mirroring operation is still live. This will kick the GC process and if the GC gets to the point of performing the leaf coalesce before the migrate completes will result in a failure of the mirror operation. So block the leaf-coalesce operation in this case. The block will be removed if another snapshot were to be taken but in the case of SXM the next expected operation is a vdi_delete of the leaf which will clear the entire chain.

### DIFF
--- a/drivers/FileSR.py
+++ b/drivers/FileSR.py
@@ -707,7 +707,7 @@ class FileVDI(VDI.VDI):
         vhdutil.killData(self.path)
 
     def _do_snapshot(self, sr_uuid, vdi_uuid, snap_type,
-                     secondary=None, cbtlog=None):
+                     _=False, secondary=None, cbtlog=None):
         # If cbt enabled, save file consistency state
         if cbtlog is not None:
             if blktap2.VDI.tap_status(self.session, vdi_uuid):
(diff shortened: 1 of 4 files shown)


drivers/FileSR.py
```diff
@@ -707,7 +707,7 @@ def reset_leaf(self, sr_uuid, vdi_uuid):
         vhdutil.killData(self.path)
 
     def _do_snapshot(self, sr_uuid, vdi_uuid, snap_type,
-                     secondary=None, cbtlog=None):
+                     _=False, secondary=None, cbtlog=None):
```
MarkSymsCtx, line 710: This just makes the method signature match the method it is overriding.

TimSmithCtx (approved):

drivers/FileSR.py
LunfanZhang: Since Python doesn’t strictly enforce matching method signatures in overrides, so maybe it`s not necessary or using `*args` could be a more flexible approach. This way, we can handle additional parameters without needing to care about their both the key and value :-)

drivers/VDI.py
```diff
@@ -16,6 +16,7 @@
 # VDI: Base class for virtual disk instances
 #
 
+import cleanup
```
rdn32, line 19: Rather than importing cleanup, maybe at some point the constants you need from cleanup could be moved into their own file.

rdn32 (approved):

drivers/VDI.py
MarkSymsCtx: Yes probably

PR #715, opened 2024-11-04
CA-400743: perform post snapshot rename in ioretry

MarkSymsCtx:
It appears that SMB3 filers or Windows hosts can return EACCES in response to a rename of a file over one of the links to a file if the rename is done too soon after the hardlink is made. Allow the rename to be retried in the event of EIO or ACCES.

### DIFF
--- a/drivers/FileSR.py
+++ b/drivers/FileSR.py
@@ -803,7 +803,10 @@ class FileVDI(VDI.VDI):
                 "FileSR_fail_snap1",
                 self.__fist_enospace)
             util.ioretry(lambda: self._snap(tmpsrc, newsrcname))
-            self._rename(tmpsrc, src)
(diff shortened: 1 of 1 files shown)


KevinLCtx (approved):

rosslagerwall:
Might this be a kernel bug? It sounds kind of similar to CA-376740 / CP-43606 although that only affects the 6.x kernel.

MarkSymsCtx:
> Might this be a kernel bug? It sounds kind of similar to CA-376740 / CP-43606 although that only affects the 6.x kernel.

I think it's a server side thing. I seem to remember reading about it somewhere.

rdn32 (approved):

rosslagerwall (approved):

TimSmithCtx (approved):

PR #717, opened 2024-11-11
CP-42675: send messages to Xapi if GC has insufficient space 

MarkSymsCtx:
Also fix an attribute error in a test flagged by static analysis

### DIFF
--- a/drivers/cleanup.py
+++ b/drivers/cleanup.py
@@ -304,6 +304,10 @@ class XAPI:
         if self.sessionPrivate:
             self.session.xenapi.session.logout()
 
+    @property
+    def srRef(self):
+        return self._srRef
+
     def isPluggedHere(self):
         pbds = self.getAttachedPBDs()
         for pbdRec in pbds:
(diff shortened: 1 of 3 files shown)


drivers/cleanup.py
```diff
@@ -1579,6 +1582,48 @@ def __init__(self, uuid, xapi, createLock, force):
...
+                    "VDI",
+                    candidate.uuid,
+                    msg_body)
+                candidate.setConfig(VDI.DB_GC_NO_SPACE, msg_id)
+
+    def clear_no_space_msg(self, vdi):
+        xapi_session = self.xapi.getSession()
+        msg_id = None
+        try:
+            msg_id = vdi.getConfig(VDI.DB_GC_NO_SPACE)
```
edwintorok, line 1618: Are you sure you want to store this per VDI? If you have 1000 VDIs, then as soon as the SR runs out of space you'd get flooded with a 1000 messages (even if the GC stops at the first one, is it guaranteed to always iterate on VDIs in the same order?)
Could this be stored per PBD or SR instead?

MarkSymsCtx: Yes, it need to be stored per VDI. As you say some may coalesce, others may not.

MarkSymsCtx: Maybe what it needs to do is raise one message to Xapi and then tag that ID to all the impacted VDIs, then remove the tag if that VDI become coalescable in the future. Finally removing the message from the SR if we complete a GC run with no VDIs unable to coalesce.

rdn32 (approved):

edwintorok (approved):

TimSmithCtx (approved):

PR #719, opened 2024-11-26
fix(cleanup.py): bad live coalesce check regarding FileSR

Wescoeur:
The `VDI.canLiveCoalesce` method can manipulates sizes of different units because of this change:
```
CP-40871: use VHD allocation size in checking canLiveCoalesce
2f863b9fce6f2978499892d8c019bb3ab7ad72c5
```
As a result, the `canLiveCoalesce` method can return `True` and cause coalesce attempts
resulting in "Timed out" exceptions.

* Only drivers deriving from `FileSR` are impacted.

* The size of `self._sizeAllocated` is calculated correctly when `vhdutil.getAllocatedSize`
is called but there is a problematic case where `getVHDInfo` is used instead.
And this function does not convert `info.sizeAllocated` from block size to bytes.

* Additionally, the getter `getAllocatedSize` should retrieve the allocated size if it's equal to -1.
Otherwise `VDI.canLiveCoalesce` can compare -1 to a positive value and would always return `True`
causing a `Timed out` error error...

### DIFF
--- a/drivers/vhdutil.py
+++ b/drivers/vhdutil.py
@@ -94,6 +94,11 @@ def ioretry(cmd, text=True):
                         errlist=[errno.EIO, errno.EAGAIN])
 
 
+def convertAllocatedSizeToBytes(size):
+    # Assume we have standard 2MB allocation blocks
+    return size * 2 * 1024 * 1024
+
+
 def getVHDInfo(path, extractUuidFunction, includeParent=True):
     """Get the VHD info. The parent info may optionally be omitted: vhd-util
     tries to verify the parent by opening it, which results in error if the VHD
@@ -115,7 +120,7 @@ def getVHDInfo(path, extractUuidFunction, includeParent=True):
             vhdInfo.parentUuid = extractUuidFunction(fields[nextIndex])
         nextIndex += 1
     vhdInfo.hidden = int(fields[nextIndex].replace("hidden: ", ""))
-    vhdInfo.sizeAllocated = int(fields[nextIndex+1])
+    vhdInfo.sizeAllocated = convertAllocatedSizeToBytes(int(fields[nextIndex+1]))
     vhdInfo.path = path
     return vhdInfo
 
@@ -274,8 +279,7 @@ def setSizePhys(path, size, debug=True):
 def getAllocatedSize(path):
     cmd = [VHD_UTIL, "query", OPT_LOG_ERR, '-a', '-n', path]
     ret = ioretry(cmd)
-    # Assume we have standard 2MB allocation blocks
-    return int(ret) * 2 * 1024 * 1024
+    return convertAllocatedSizeToBytes(int(ret))
 
 def killData(path):
     "zero out the disk (kill all data inside the VHD file)"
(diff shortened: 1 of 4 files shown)


MarkSymsCtx:
Could you please add some Unit Tests to `tests/test_vhdutil.py` to regression control this change.

drivers/vhdutil.py
```diff
@@ -115,7 +120,7 @@ def getVHDInfo(path, extractUuidFunction, includeParent=True):
             vhdInfo.parentUuid = extractUuidFunction(fields[nextIndex])
         nextIndex += 1
     vhdInfo.hidden = int(fields[nextIndex].replace("hidden: ", ""))
-    vhdInfo.sizeAllocated = int(fields[nextIndex+1])
+    vhdInfo.sizeAllocated = convertAllocatedSizeToBytes(int(fields[nextIndex+1]))
```
LunfanZhang, line 123: is the block size a fixed value here?

MarkSymsCtx: it technically is not but in practice all VHDs created in XenServer are using 2MB allocation blocks. The allocation block size is a property in the VHD header.

LunfanZhang: > it technically is not but in practice all VHDs created in XenServer are using 2MB allocation blocks.
Thanks for point this out @MarkSymsCtx , then fixed value work.
from a compatibility perspective, I think calculating it with size in VHD metadata would be better. 
Just a thought for future improvements.

MarkSymsCtx:
> We just added some tests regarding vhdutil, can you confirm everything is ok? Thanks!

We'd prefer the block size to be read from the VHD header rather than hard coded to 2M

olivierlambert:
Hi Mark,

Thank you for your feedback. I’d like to address a few points to improve collaboration in this open-source project.

This PR fixes a regression introduced in a prior commit you authored. While we all make mistakes, an acknowledgment of the urgency and impact of this fix on downstream projects would have been appreciated. Such recognition fosters better collaboration.

When we submitted the PR, you requested additional unit tests, which we promptly provided within two days. However, there was no feedback or action for two weeks, delaying the resolution of this critical issue. After that period, you raised a new concern about reading the block size from the VHD header rather than hard-coding it. While valid, earlier feedback would have avoided unnecessary delays.

**Given the urgency of this bug, perhaps we could address the dynamic block size in a separate PR, allowing us to prioritize the immediate fix. This approach would resolve the regression more quickly while still enabling improvements in subsequent work.** What do you think?

As contributors to a Linux Foundation-hosted project, we dedicate significant effort to improving the toolstack for all users. Constructive feedback is always welcome, but acknowledgment of contributors' efforts—especially when fixing regressions—is essential for maintaining goodwill and motivation.

I hope we can move forward collaboratively. Please let us know if there are any further requests for this PR.

MarkSymsCtx (approved):

KevinLCtx (approved):

olivierlambert:
Thank you very much for merging! As promised, we'll provide improvements via an extra PR in January. Until then, we can provide the fix downstream for everyone :+1:

PR #722, opened 2024-12-09
CP-52852: add handler for xmlrpc ProtocolError

MarkSymsCtx:
It is possible that a XenAPI call to xapi will result in an xmlrpc ProtocolError being returned, typically a 500 Internal Error. As it is not practical to wrap all the calls to xapi via the xenapi session in a specific exception handler add one into the outer run method to ensure that it is converted to an error code and not an unknown error.

### DIFF
--- a/drivers/SRCommand.py
+++ b/drivers/SRCommand.py
@@ -17,11 +17,13 @@
 #
 # SRCommand: parse SR command-line objects
 #
+import traceback
 
 import XenAPI # pylint: disable=import-error
 import sys
 import xs_errors
 import xmlrpc.client
+from xmlrpc.client import ProtocolError
 import SR
 import util
 import blktap2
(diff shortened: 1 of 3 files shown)


KevinLCtx (dismissed):

drivers/SRCommand.py
```diff
@@ -390,6 +392,12 @@ def run(driver, driver_info):
         else:
             print(ret)
 
+    except ProtocolError:
+        # xs_errors.__new__ returns a different class
+        # pylint: disable-msg=E1101
+        print(xs_errors.XenError(
```
edwintorok, line 398: You may not have anyone to report to, e.g. if someone did a toolstack restart and get EPIPE on this probably.
Can you also log it to a file like with other exceptions?

MarkSymsCtx: Done

KevinLCtx (approved):

edwintorok (approved):

robhoes (approved):

PR #731, opened 2025-01-27
CA-405417: add tracing of process holding device

MarkSymsCtx:


drivers/lvutil.py
```diff
@@ -446,15 +446,15 @@ def _openExclusive(dev, retry):
     try:
         return os.open("%s" % dev, os.O_RDWR | os.O_EXCL)
     except OSError as ose:
-        opened_by = ''
+        ret = util.pread2(["lsof", dev])
```
edwintorok, line 449: I think `lsof` can get stuck though if you have a stuck remote filesystem (NFS/GFS2), because as it looks through `/proc` it might try to resolve links pointing to these FS.

Or at least `fuser` would get stuck, and we had to remove a usage of that from XAPI to prevent XAPI from getting stuck:
https://github.com/xapi-project/xen-api/pull/6197

Is it possible to run this command, but with a timeout so it doesn't completely block everything?

MarkSymsCtx:
As we've found issues with fuser and lsof getting stuck when unrelated SRs are in an unresponsive state, declining this.

PR #734, opened 2025-03-03
CA-407566: adjust systemd dependencies

MarkSymsCtx:
The previous set of dependencies actually result in an ordering loop and aren't actually required.

### DIFF
--- a/systemd/sm-mpath-root.service
+++ b/systemd/sm-mpath-root.service
@@ -1,8 +1,5 @@
 [Unit]
 Description=Ensure symlinks for multipathed rootdev
-Wants=basic.target multipathd.service
-After=basic.target
-Before=multipathd.service
 
 [Service]
 Type=oneshot
@@ -10,4 +7,4 @@ RemainAfterExit=yes
 ExecStart=/opt/xensource/sm/multipath-root-setup
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=sysinit.target


rosslagerwall (approved):

systemd/sm-mpath-root.service
```diff
@@ -1,13 +1,10 @@
...
-Before=multipathd.service
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/opt/xensource/sm/multipath-root-setup
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=sysinit.target
```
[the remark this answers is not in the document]
MarkSymsCtx: I think so long as it's early and sysinit should be early enough that should be fine as the SM can't start until after xapi.

TimSmithCtx, line 10: According to ```man bootup``` on my system, ```sysinit.target``` occurs strictly before ```basic.target```, and according to ```man systemd.special```

```
       basic.target
           A special target unit covering basic boot-up.

           systemd automatically adds dependency of the type After= for this target unit to all services (except for those with DefaultDependencies=no).
```

So do we need to put ```DefaultDependencies=no``` in here?

MarkSymsCtx: it doesn't seem so. But we should probably perform a separate system wide check of the startup sequencing before GA.

TimSmithCtx (approved):

PR #740, opened 2025-03-18
Begin moving SM python library code into proper python install paths

TimSmithCtx:
The SM code should not be kept in /opt but should instead live in the proper python install paths (if it is library code) or under /usr/libexec or /usr/bin if it is not.

A previous attempt was made to start putting library code under proper python install paths in the form of ```sm-core-libs```, but this approach made maintenance difficult and introduced divergence.

This creates a new sub-package of SM which will install as ```python<version>-sm-libs``` which obsoletes and replaces ```sm-core-libs``` and begins the process of moving the existing SM library code into the proper python install paths, while leaving wrappers behind in /opt/xensource/sm so that code relying on older names and/or paths can continue to work. Eventually all of /opt/xensource/sm will be in a compatibility sub-package which can then be dropped in a major version update at some point.

Since there are few consumers of ```sm-core-libs``` the focus of resolution is to provide its functionality in terms of ```sm-libs``` making as little change to the SM code as possible, beyond moving it. We can then begin making the interface more rational and updating present consumers of ```sm-core-libs``` to use it.


TimSmithCtx:
Hmm. Python 3.11 likes this, but python 3.6 doesn't...

TimSmithCtx:
> Hmm. Python 3.11 likes this, but python 3.6 doesn't...

I think I need __init__.py in the source, rather than created in the Makefile

TimSmithCtx:
...and of course the workflow file needed an update, to add the new libs location.

tests/test_cbt.py
```diff
@@ -397,28 +397,19 @@ def test_snapshot_success_with_parent(self, context, mock_lock,
...
-            mock_ensure_space, mock_logcheck, mock_cbt):
+    def test_snapshot_out_of_space_failure(self, mock_ensure_space, mock_logcheck, mock_cbt):
         # Create the test object
         self.vdi = TestVDI(self.sr, self.vdi_uuid)
         # Set initial state
         self._set_initial_state(self.vdi, True)
         snap_uuid = uuid.uuid4()
 
-        # Write util.SMlog calls to smlog_out
-        self.fakesmlog = ""
```
MarkSymsCtx, line 410: Can this not still be done? It seems to be checking an important side-effect of the method call, i.e. that a log message is emitted to say there is insufficient space.

Makefile
```diff
@@ -14,36 +15,58 @@ SM_DRIVERS += LVHDoHBA
...
 SM_LIBS += vhdutil
 SM_LIBS += lvhdutil
 SM_LIBS += cifutils
-SM_LIBS += xs_errors
 SM_LIBS += nfs
 SM_LIBS += devscan
 SM_LIBS += sysdevice
-SM_LIBS += iscsilib
-SM_LIBS += mpath_dmp
 SM_LIBS += mpath_null
```
MarkSymsCtx, line 63: Should this (mpath_null) not move as well?

tests/test_cbt.py
TimSmithCtx: The log message was a side-effect of creating a XenError object, which seemed like the wrong place to be logging from, and with the compatibility layers ended up creating a circular import failure because ```xs_errors``` imported ```util```, which imported ```xs_errors```...

I could take a look at fixing that by adding the logging to ```drivers/VDI.py``` where the exception is handled, or alternatively it would be fixable by reinstating the logging side-effect from ```xs_errors.XenError()``` and moving all the logging code out of ```util``` and into a new library, which is a bigger change.

MarkSymsCtx: OK, that sounds fair.

MarkSymsCtx (approved):

KevinLCtx (approved):

PR #742, opened 2025-03-18
CP-53050: drop multipath.conf

MarkSymsCtx:
This is now owned and installed by device-mapper-multipath.


MarkSymsCtx:
This has been pending for a while since we stopped installing it.

LunfanZhang (approved):

PR #743, opened 2025-03-19
EXT: don't reserve any block for root

ydirson:
There is only root in dom0, and `df` erroneously substracts the default 5% of disk space as "reserved for root", which is just meaningless here.

It could be worth going further with doing the same for all FileSR types, but I thought I'd ask for feedback first.

If this becomes a thing, it could be useful then to find some way to get `tune2fs -m 0` run on existing SR to make them even.

### DIFF
--- a/drivers/EXTSR.py
+++ b/drivers/EXTSR.py
@@ -205,7 +205,7 @@ class EXTSR(FileSR.FileSR):
                 opterr='Insufficient space in VG %s' % self.vgname)
 
         try:
-            util.pread2(["mkfs.ext4", "-F", self.remotepath])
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
As you say, all access is performed as root anyway so I'm not sure what actual difference this will make?

rosslagerwall:
> There is only root in dom0

This assertion is not correct for XenServer. In XenServer's dom0 various processes run as different users (e.g. chrony, pvsproxy, QEMU, swtpm, ...). A vulnerability without a privilege escalation in one of those components could then more easily lead to a DoS if the reserve is removed, i.e. it would be a loss of defence-in-depth. Whether this is important enough to keep the reserve is debatable but the discussion at least needs to be had.

PR #744, opened 2025-03-20
CP-53692 SR attach calls mpathcount async

LunfanZhang:
Kickpipe as subprocess to run mpathcount script in the background to avoid the synchronously calling block the SR attach process.

The original implementation will pass a `SCSIid` to the mpathcount script, however I think this script will not handler a extra parameter at the moment. so it is an unnecessary parameter.

BST - 214740 

Limit configure test -  21474

### DIFF
--- a/drivers/SR.py
+++ b/drivers/SR.py
@@ -496,8 +496,7 @@ class SR(object):
             self.session.xenapi.SR.set_sm_config(self.sr_ref, sm_config)
 
             if self.mpath == "true" and len(SCSIid):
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
I think it's fine that mpathcount no longer takes a parameter, I think we did away with that when it was converted to a kickpipe service.

MarkSymsCtx (approved):

KevinLCtx (approved):

TimSmithCtx (approved):

PR #745, opened 2025-03-27
More more of SM out of /opt/xensource

TimSmithCtx:
This is another installment of moving parts of SM out of /opt/xensource and into more correct locations.
Scripts and executables should go in /usr/libexec/sm while python library code belongs under the proper python include paths.

There is much yet to do, but to keep the individual changes smaller, sending this batch for review now.


tests/test_BaseISCSI.py
```diff
@@ -85,6 +86,7 @@ def test_attach_tgt_present_path_found(self, mock_init_paths):
         self.subject.attach(self.sr_uuid)
 
     @mock.patch('BaseISCSI.BaseISCSISR._initPaths', autospec=True)
+    @mock.patch('sm.core.xs_errors.XML_DEFS', 'libs/sm/core/XE_SR_ERRORCODES.xml')
```
MarkSymsCtx, line 89: This shouldn't be required if it's done at the class level?

tests/test_FileSR.py
```diff
@@ -438,6 +439,9 @@ class TestShareFileSR(unittest.TestCase):
     ERROR_524 = "Unknown error 524"
 
     def setUp(self):
+        xml_patcher = mock.patch('sm.core.xs_errors.XML_DEFS','libs/sm/core/XE_SR_ERRORCODES.xml')
```
MarkSymsCtx, line 442: If we have this mocked at the class level this bit in `setUp` should not be required?

MarkSymsCtx (approved):
Couple of minor comments about possibly unnecessary patching of the error defs in the tests but that can be cleaned up later.

MarkSymsCtx:
You do need to sign the commits off though

tests/test_BaseISCSI.py
TimSmithCtx: Ah missed some removals of an interim attempt.

tests/test_FileSR.py
TimSmithCtx: This one wasn't mocked at the class level. It is now though.

MarkSymsCtx (approved):

drivers/mpath_null.py
```diff
@@ -1,36 +1 @@
...
-    return
-
-
-def deactivate():
-    return
-
-
-def path(SCSIid):
-    return DEVBYIDPATH + "/scsi-" + SCSIid
+from sm.core.mpath_null import *
```
LunfanZhang, line 1: Since we have removed all the content from this file, why this line is still needed?

TimSmithCtx: Like the others with similar content, it is there in case there is some external code which manipulates the python include path and tries to import it using its old location in ```/opt/xensource/sm```. We simply cannot know what might be out there and relying on it.

Because _everything_ which imported from ```/opt/xensource/sm ``` does so by manipulating the python include path (because that is not one of the standard import locations), moving the install location _without_ leaving this compatibility file behind would be a change of interface, which would require a major version update which we are not yet ready to do.

LunfanZhang (approved):

PR #747, opened 2025-04-03
CP-54096 Further evacuation of /opt

TimSmithCtx:
Move the cleanup and mpathcount executables from /opt

Also clean up some references to /bin which should now be using /usr/bin


MarkSymsCtx:
you've got a couple of pylint complaints
```
************* Module mpathcount
drivers/mpathcount:85:39: E0606: Possibly using variable 'mpath_enabled' before assignment (possibly-used-before-assignment)
************* Module cleanup
drivers/cleanup:56:4: E0606: Possibly using variable 'cleanup' before assignment (possibly-used-before-assignment)
```
you might need to just mark them with appropriate comments if they're false positives

drivers/cleanup
```diff
@@ -0,0 +1,137 @@
...
+        usage()
+    if action == "debug" and not (debug_cmd and vdi_uuid) or \
+            action != "debug" and (debug_cmd or vdi_uuid):
+        usage()
+
+    if action != "query" and action != "debug":
+        print("All output goes to log")
+
+    if action == "gc":
+        cleanup.gc(None, uuid, background, dryRun)
```
MarkSymsCtx, line 119: need a `from sm import cleanup`?

MarkSymsCtx (changes requested):

drivers/mpathcount
```diff
@@ -0,0 +1,94 @@
...
+    localhost = session.xenapi.host.get_by_uuid(mpathcount.get_localhost_uuid())
+    mpathcount.check_xapi_is_enabled()
+    # Check whether multipathing is enabled (either for root dev or SRs)
+    try:
+        if mpathcount.get_root_dev_major() != mpathcount.get_dm_major():
+            hconf = session.xenapi.host.get_other_config(localhost)
+            assert(hconf['multipathing'] == 'true')
+            mpath_enabled = True
+    except:
+        mpath_enabled = False
```
MarkSymsCtx, line 45: This should be initialised to False  above the try as the `if` might evaluate False

TimSmithCtx:
I'm looking into why the "make precheck" in the check job here looks so different from when I run it locally.

MarkSymsCtx:
> I'm looking into why the "make precheck" in the check job here looks so different from when I run it locally.

Probably a subtly different version of pylint

drivers/cleanup
TimSmithCtx: There is one below, which will always have executed by the time this is called, but I'll move it up to the top.

drivers/mpathcount
TimSmithCtx: No, this one should be addressed as mpathcount.mpath_enabled...

drivers/cleanup
```diff
@@ -0,0 +1,137 @@
...
+    elif action == "query":
+        print("Currently running: %s" % cleanup.get_state(uuid))
+    elif action == "disable":
+        cleanup.abort_optional_reenable(uuid)
+    elif action == "debug":
+        cleanup.debug(uuid, debug_cmd, vdi_uuid)
+
+
+if __name__ == '__main__':
+    sys.path.append("/opt/xensource/sm/")
```
MarkSymsCtx, line 135: I don't think you need the path override here?

drivers/intellicache-clean
```diff
@@ -21,7 +21,7 @@ import sys
 import XenAPIPlugin
 sys.path.append("/opt/xensource/sm/")
```
MarkSymsCtx, line 22: Path override probably no longer required?

drivers/cleanup
TimSmithCtx: It needs to be present, because ```from sm import cleanup``` causes other imports which are still in /opt/xensource/sm for the moment, as it says in the commit message for that change.

It (and the subsequent import) do need to go up to the top of the file to resolve the pylint complaint.

drivers/intellicache-clean
TimSmithCtx: Do though, for the same reason as above. Can drop these when everything that gets imported no longer references anything in the compat path.

(this is why ```/usr/libexec/sm/mpathcount``` _doesn't_ need it)

KevinLCtx (approved):

PR #752, opened 2025-04-11
Move remaining libraries out of /opt

TimSmithCtx:
This clears /opt/xensource/sm of non-except for drivers, and hybrids which are both libraries and have __main__ sections and need to be separated.


libs/sm/LUNperVDI.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/python3
```
MarkSymsCtx, line 1: This should loose the shebang as it's not executable. Basically, general comment, nothing in `libs/sm` should have a shebang

libs/sm/BaseISCSI.py
```diff
@@ -0,0 +1,767 @@
+#!/usr/bin/python3
```
MarkSymsCtx, line 1: This should loose the shebang as it's not executable.

libs/sm/LUNperVDI.py
TimSmithCtx: Moved them with "git mv". I'll add a commit to the series.

MarkSymsCtx: > Moved them with "git mv". I'll add a commit to the series.

Annoying that Github doesn't show them like that then :(

TimSmithCtx: That's because of the breadcrumb still left in for backwards compatibility. It confuses git a bit unfortunately, but I wanted to keep each commit bisectable in case of issues.

MarkSymsCtx (approved):

Makefile
```diff
@@ -32,7 +32,9 @@ SM_CORE_LIBS += f_exceptions
 # sm-core-libs provided differently.
 SM_CORE_LIBS += libiscsi
 
-SM_LIBS := cleanup
+SM_LIBS :=
```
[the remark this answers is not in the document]
TimSmithCtx: But then, when adding a new entry using ```+=```, it becomes fiddly to filter the block through ```sort -u``` to make the subsequent diffs easy to read because new lines are inserted in lexical order. 

Leaving the first line as ```SM_LIBS :=``` means it always sorts first and keeps everything working nicely.

PR #753, opened 2025-04-22
CP-54434 Move remaining executables from /opt

TimSmithCtx:


drivers/blktap2
```diff
@@ -0,0 +1,174 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+import sys
+sys.path.append("/opt/xensource/sm/")
```
MarkSymsCtx, line 18: Still to be removed?

MarkSymsCtx (approved):
Still more sys.path manipulation than I was expecting at this stage.

drivers/blktap2
TimSmithCtx: Yes. There will need to be a pass when all is done to get rid of all of these.

libs/sm/SR.py
```diff
@@ -469,13 +469,20 @@ def _mpathinit(self):
                 self.mpath = "false"
                 self.mpathhandle = "null"
 
-            if not os.path.exists("/opt/xensource/sm/mpath_%s.py" % self.mpathhandle):
-                raise IOError("File does not exist = %s" % self.mpathhandle)
+            # Attempt to import the specified mpath module and return None
+            # if this does not throw an exception.
+            module_name = "sm.core.mpath_%s" % self.mpathhandle
+            self.mpathmodule = __import__(module_name, fromlist=[None])
+            return None
```
LunfanZhang, line 476: The function would still implicitly `return None` when it reaches the end of execution, so why we need return it explicitly here. line 485 as well

LunfanZhang (approved):

libs/sm/SR.py
TimSmithCtx: Because an explicit return clearly indicates an intended behaviour to someone editing the code in the future, as well as to future versions of the interpreter which might choose to alter the implicit behaviour (python has a track record of doing this).

Allowing the implicit behaviour to provide the return does not.

PR #755, opened 2025-05-01
CP-54766 Move SM SR drivers from /opt

TimSmithCtx:
This involves splitting a number of files which were used as both imported libraries and direct executables into parts, and since git tracks file-renames by inference using content-similarity sometimes it shows a rename+diff which is not correct.


TimSmithCtx:
Annoyingly, forgot the signoff the first time...

drivers/NFSSR
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python3
+#
+# Copyright (C) Citrix Systems Inc.
```
LunfanZhang, line 3: Is this the copyright we using now, or the CSG?

LunfanZhang (approved):

PR #766, opened 2025-07-30
CA-414438 Defend against unexpected VDI name_description contents

TimSmithCtx:
It seems to be possible for a description field on an LV snapshot to be converted to a dictionary, which results in attempts to remove it from the MGT volume failing after it is deleted, which in turn results in a subsequent scan failing because an LV which should exist does not.

It is not immediately clear how this ended up as a dictionary as it has been back and forth through XMLRPC a few times, but to solve the immediate issue, defend against such things.


libs/sm/core/util.py
```diff
@@ -118,6 +118,8 @@ def to_plain_string(obj):
         return None
     if type(obj) == str:
         return obj
+    if type(obj) == dict and len(obj) == 0:
```
LunfanZhang, line 121: Following Python best practice, `isinstance(obj, dict)` is preferred over `type(obj)`, but both are worked.

LunfanZhang (approved):

libs/sm/core/util.py
MarkSymsCtx: Might as well tidy this up while we're making changes

MarkSymsCtx: As we don't expect this to be the data type that we receive from xapi I wonder if we should log what we got as it might then allow for determining what went wrong?

MarkSymsCtx (approved):

libs/sm/core/util.py
MarkSymsCtx: Dangling bits of py2 -> py3 I suspect. It probably had basestring and unicode in there at some point.

PR #770, opened 2025-08-21
fix: stop SMGC service on SR detach to prevent orphaned systemd units

Millefeuille42:
This is not done on every and each implementation of SR but only on ones that calls `cleanup.start_gc_service` (like `FileSR`) and on the classes that inherits from them and don't call super on `detach`.

This is to prevent useless errors logs like `Failed to stop xxx.service: Unit xxx.service not loaded.`

### DIFF
--- a/libs/sm/cleanup.py
+++ b/libs/sm/cleanup.py
@@ -3205,6 +3205,7 @@ def abort(srUuid, soft=False):
     """Abort GC/coalesce if we are currently GC'ing or coalescing a VDI pair.
     """
     if _abort(srUuid, soft):
+        stop_gc_service(srUuid)
         Util.log("abort: releasing the process lock")
         lockGCActive.release()
         return True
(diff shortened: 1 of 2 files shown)


libs/sm/drivers/LVHDSR.py
```diff
@@ -641,6 +641,7 @@ def detach(self, uuid):
         # However, we should still delete lock files on slaves as it is the
         # only place to do so.
         self._cleanup(self.isMaster)
+        cleanup.stop_gc_service(sr_uuid)
```
MarkSymsCtx, line 644: This should replace the `cleanup.abort()` at line 589, or alternatively make `cleanup.abort()` stop the service.

MarkSymsCtx (changes requested):
Can be significantly simplified by making `cleanup.abort` perform this operation

MarkSymsCtx (approved):

PR #775, opened 2025-09-16
CA-405215: attach tapdisk pid to correct cgroup path

chunjiez:
it works on both cgroup v1 and v2


TimSmithCtx:
This should be fixed in the cgrules.conf rather than here.

PR #797, opened 2026-01-26
Revert baa7044

MarkSymsCtx:
The commit baa7044 introduced regressions in functionality around the use of `raw` VDIs on LVM SRs. It is therefore being reverted until it can be reworked to not cause these regressions.

@Millefeuille42 I will reopen your original PR, can you please pick it back up and address the issues? Possibly something in the `block_size` method needing to check the VDI type before calling `vhdutil`. In the failures that we've seen we have traces where we can see that `vhdutil` has tried, unsuccessfully, to read a `raw` VDI and thus caused the outer operation to fail. Disaster Recovery functionality was affected as well as various positive and negative tests around use of `raw` VDIs

### DIFF
--- a/libs/sm/VDI.py
+++ b/libs/sm/VDI.py
@@ -101,7 +101,6 @@ class VDI(object):
         self.description = ''
         self.vbds = []
         self.size = 0
-        self._block_size = -1
         self.utilisation = 0
         self.vdi_type = ''
         self.has_child = 0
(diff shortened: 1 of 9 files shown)


LunfanZhang (approved):

TimSmithCtx (approved):

Millefeuille42:
The fix have been made and is pushed on our branch [xcp-ng:upstream-use-vhd-header-for-block-size](https://github.com/xcp-ng/sm/tree/upstream-use-vhd-header-for-block-size) in commit https://github.com/xapi-project/sm/commit/9ce01d1bff62d2e0ebc160a5a5965f6ff5facf22.

It should be good for review once the PR is back open. I managed to reproduce the issue and tested the fix on a raw VDI on a LVM SR. I managed to run several commands (create, resize, scan, destroy...) without any issue.

